PR #2, opened 2019-04-09
CDES-1033: changed back end local port

mattclark-zerogravit:


conf/application.conf
```diff
@@ -151,10 +142,10 @@ contact-frontend {
...
 
 mongodb {
-  uri = "mongodb://localhost:27017/customs-declare-exports-frontend"
+  uri = "mongodb://localhost:27017/customs-movements-frontend"
 }
 
 urls {
   login = "http://localhost:9949/auth-login-stub/gg-sign-in"
-  loginContinue = "http://localhost:9000/customs-declare-exports-frontend"
+  loginContinue = "http://localhost:9000/customs-movements-frontend"
```
PatrykRudnicki: Are you sure about port? I think we should use localhost with port for movements frontend

mattclark-zerogravit: you are right. missed that! will sort !

PR #3, opened 2019-04-16
CEDS-1056 Consignment reference page with refactor

PatrykRudnicki:
Deprecated for CustomsBaseSpec because I want to use cleaner approach, It's difficult to remove and switch to use new implementation now and actual task will cover all screens so it's only information for me to not using this, at the end it'll be removed. This task will be covered in few PRs, so don't worry about it.

Coverage changed to 80 %, because I don't want to spend time to add tests for actual implementation. Changes that I added are covered in 100 %.

The same with build.sbt changes, only to allow me use deprecated, it'll come back when I finish task.

Second commit contains only scalafmt changes, I prefer to do it in one commit and one place, not running scalafmt only for new files.

NRSConnector is removed after talk with @raghu1978. We will implement it in separate repo and moreover we didn't have usage for it.

This PR doesn't contain view spec. I'll add it at the end, probably there will be some changes around screens. Easier approach for me.


app/forms/MovementForms.scala
```diff
@@ -112,13 +112,13 @@ object Movement {
   def createMovementRequest(cacheMap: CacheMap,
                             eori: String,
                             choice: Choice): InventoryLinkingMovementRequest = {
-    val ducrForm = cacheMap.getEntry[Ducr](Ducr.id).get
+    val ducrForm = cacheMap.getEntry[Ducr](Ducr.id).getOrElse(Ducr(""))
     val goodsDate =
       cacheMap.getEntry[GoodsDateForm](MovementFormsAndIds.goodsDateId)
     val location =
-      cacheMap.getEntry[LocationForm](MovementFormsAndIds.locationId).get
+      cacheMap.getEntry[LocationForm](MovementFormsAndIds.locationId).getOrElse(LocationForm(None, None, None, None))
```
raghu1978: didn't understand the requirement of creating a form if None of the fields are populated.

app/forms/MovementForms.scala
```diff
@@ -132,14 +132,14 @@ object Movement {
                      agentLocation = location.agentLocation,
                      agentRole = location.agentRole)),
       ucrBlock = UcrBlock(ucr = ducrForm.ducr, ucrType = "D"),
-      goodsLocation = location.goodsLocation.get,
+      goodsLocation = location.goodsLocation.getOrElse(""),
       goodsArrivalDateTime =
         if (choice.value.equals("EAL") && goodsDate.isDefined)
-          Some(extractDateTime(goodsDate.get))
+          Some(extractDateTime(goodsDate.getOrElse(GoodsDateForm("", "", "", None, None))))
```
raghu1978: same comment as above

PR #19, opened 2019-07-04
CEDS-1219 Finished updating play framework - removed and replaced deprecated

PatrykRudnicki:


app/handlers/ErrorHandler.scala
```diff
@@ -31,16 +31,18 @@ import uk.gov.hmrc.play.bootstrap.http.FrontendErrorHandler
 import scala.concurrent.Future
 
 @Singleton
-class ErrorHandler @Inject()(appConfig: AppConfig, val messagesApi: MessagesApi)
-    extends FrontendErrorHandler with I18nSupport with AuthRedirects {
+class ErrorHandler @Inject()(
+  appConfig: AppConfig,
+  val messagesApi: MessagesApi,
+  errorTemplate: views.html.error_template
```
ed0906: maybe import here rather than `views.html.`

PR #20, opened 2019-07-04
CEDS-1257 Review Arrive & Depart Validation

PauloPortugal:
* Validate EORI and MUCR/DUCR
* Validate two letter country codes

### DIFF
--- a/app/utils/validators/forms/FieldValidator.scala
+++ b/app/utils/validators/forms/FieldValidator.scala
@@ -44,6 +44,12 @@ object FieldValidator {
     def or(second: Boolean): Boolean = first || second
   }
 
+  val validEori: String => Boolean = (input: String) => input.matches("^GB[0-9]{1,15}$")
+  val validDucrOrMucr: String => Boolean = (input: String) => input.matches("""^[0-9][A-Z][A-Z][0-9A-Z\(\)\-/]{6,32}|
+                                                                         |GB/[0-9A-Z]{3,4}-[0-9A-Z]{5,28}|
+                                                                         |GB/[0-9A-Z]{9,12}-[0-9A-Z]{1,23}|
+                                                                         |A:[0-9A-Z]{3}[0-9]{8}|
+                                                                         |C:[A-Z]{3}[0-9A-Z]{3,30}""")
   private val zerosOnlyRegexValue = "[0]+"
   private val noMoreDecimalPlacesThanRegexValue: Int => String =
     (decimalPlaces: Int) => s"^([0-9]*)([\\.]{0,1}[0-9]{0,$decimalPlaces})$$"
--- a/app/utils/validators/forms/FieldValidator.scala
+++ b/app/utils/validators/forms/FieldValidator.scala
@@ -16,6 +16,8 @@
 
 package utils.validators.forms
 
+import services.Countries.allCountries
+
 import scala.util.{Success, Try}
 
 object FieldValidator {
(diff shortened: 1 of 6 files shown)


test/forms/ConsignmentReferencesSpec.scala
```diff
@@ -42,28 +42,46 @@ class ConsignmentReferencesSpec extends BaseSpec {
...
     }
 
     "return error for incorrect reference" in {
-      val inputData = ConsignmentReferences(Some("GB71757250450281160"), "Incorrect reference", "12345")
+      val inputData =
+        ConsignmentReferences(Some("GB71757250450281160"), "Incorrect reference", "5123456789-000-123ABC45$%^FIIIII")
       val errors = ConsignmentReferences.form().fillAndValidate(inputData).errors
 
-      errors.length must be(2)
+      errors.length must be(3)
```
benjaminparker: Assert this in a single line.  errors mustBe List(....)

test/forms/ConsignmentReferencesSpec.scala
```diff
@@ -42,28 +42,46 @@ class ConsignmentReferencesSpec extends BaseSpec {
...
       errors(1) must be(FormError("reference", "consignmentReferences.reference.error"))
+      errors(2) must be(FormError("referenceValue", "consignmentReferences.reference.value.error"))
     }
 
-    "no errors when data is correct" in {
-      val inputData = ConsignmentReferences(Some("GB717572504502811"), "D", "123456")
-      val errors = ConsignmentReferences.form().fillAndValidate(inputData).errors
+    "no errors when data is correct with different valid MUCR/DUCRs" in {
+      val inputData1 = ConsignmentReferences(Some("GB717572504502811"), "D", "5GB123456789000-123ABC456DEFIIIII")
+      val errors1 = ConsignmentReferences.form().fillAndValidate(inputData1).errors
```
benjaminparker: use mustBe empty

test/forms/ConsignmentReferencesSpec.scala
```diff
@@ -42,28 +42,46 @@ class ConsignmentReferencesSpec extends BaseSpec {
...
+      errors1.length must be(0)
+
+      val inputData2 = ConsignmentReferences(Some("GB717572504502811"), "D", "GB/ABC4-ASIUDYFAHSDJF")
+      val errors2 = ConsignmentReferences.form().fillAndValidate(inputData2).errors
+      errors2.length must be(0)
```
benjaminparker: use mustBe empty

app/utils/validators/forms/FieldValidator.scala
```diff
@@ -44,6 +46,12 @@ object FieldValidator {
     def or(second: Boolean): Boolean = first || second
   }
 
+  val validEori: String => Boolean = (input: String) => input.matches("^GB[0-9]{1,15}$")
```
benjaminparker: We could use tiny types like case class Eori(val: String) extends AnyVal.  Then put the valid Regex with the class.  Then we don't have the possibility of another String ending up there and we have the validation with the thing being validated.  An easy fix that would save a lot of bugs ?

PauloPortugal: I am a fan of dynamic typed functional languages :)
This function solves the problem at hand.

test/forms/ConsignmentReferencesSpec.scala
PauloPortugal: Addressed!

test/forms/ConsignmentReferencesSpec.scala
PauloPortugal: Addressed!

test/forms/ConsignmentReferencesSpec.scala
PauloPortugal: Addressed!

PR #22, opened 2019-07-09
CEDS-1204 Clean up IntelliJ Errors

ed0906:
IntelliJ struggles to render templates properly when there are two packages/classes imported with the same name.

For example twirl imports a `utils` package by default (part of Play). If we then import our own `utils` package the template compiles, but IntelliJ gives a tonne of errors. 

Given the fix was quick and easy I figured it was worth doing

### DIFF
--- a/app/views/arrival_details.scala.html
+++ b/app/views/arrival_details.scala.html
@@ -30,7 +30,7 @@
     @helper.form(routes.MovementDetailsController.saveMovementDetails(), 'autoComplete -> "off") {
         @helper.CSRF.formField
 
-        @components.back_link("/customs-movements/consignment-references")
+        @components.back_link(routes.ConsignmentReferencesController.displayPage())
 
         @components.error_summary(form.errors)
 
--- a/app/views/arrival_details.scala.html
+++ b/app/views/arrival_details.scala.html
@@ -15,7 +15,7 @@
  *@
 
 @import config.AppConfig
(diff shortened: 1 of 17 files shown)


app/views/components/RadioOption.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package utils
+package views.components
```
MaciejRewera: This is not a view component. This is why it was in `utils`.

app/views/arrival_details.scala.html
```diff
@@ -27,7 +27,7 @@
     title = messages("arrivalDetails.title"),
     appConfig = appConfig
 ) {
-    @helper.form(MovementDetailsController.saveMovementDetails(), 'autoComplete -> "off") {
+    @helper.form(routes.MovementDetailsController.saveMovementDetails(), 'autoComplete -> "off") {
```
MaciejRewera: 👍

app/views/components/RadioOption.scala
ed0906: Agreed `components` might as well be called `utils` given everything that is in it. I just did the quickest thing to get it to compile

ed0906: I'd argue that "helpers that are only used by views" should go in the `views` package

PR #24, opened 2019-07-09
CEDS-1257 Allow other countries to be entered on EORI

PauloPortugal:
Update the countries code list to include Kosovo and Saint Barthélemy.
We also need to review some of these "countries" as they are more
regions.


test/forms/TransportSpec.scala
```diff
@@ -84,10 +84,15 @@ class TransportSpec extends BaseSpec {
 
       "values are correct" in {
 
-        val inputData = Transport(Sea, "PL")
-        val errors = Transport.form().fillAndValidate(inputData).errors
+        val inputData1 = Transport(Sea, "PL")
+        val errors1 = Transport.form().fillAndValidate(inputData1).errors
```
benjaminparker:
```suggestion
        Transport.form().fillAndValidate(inputData1).errors should be (empty) 
```

test/forms/TransportSpec.scala
```diff
@@ -84,10 +84,15 @@ class TransportSpec extends BaseSpec {
 
       "values are correct" in {
```
benjaminparker: can we describe this a little more in domain terms?

test/forms/TransportSpec.scala
```diff
@@ -84,10 +84,15 @@ class TransportSpec extends BaseSpec {
...
 
-        val inputData = Transport(Sea, "PL")
-        val errors = Transport.form().fillAndValidate(inputData).errors
+        val inputData1 = Transport(Sea, "PL")
```
benjaminparker: Call this by a more specific name.  Like transportPoland ?

test/forms/TransportSpec.scala
```diff
@@ -84,10 +84,15 @@ class TransportSpec extends BaseSpec {
...
+        val errors1 = Transport.form().fillAndValidate(inputData1).errors
+
+        errors1.length must be(0)
+
+        val inputData2 = Transport(Sea, "XK")
```
benjaminparker: Move this to a separate assertion and then we can check each A/C separately.  Also avoids the inputData2 val name.

test/forms/TransportSpec.scala
```diff
@@ -84,10 +84,15 @@ class TransportSpec extends BaseSpec {
...
+        val errors2 = Transport.form().fillAndValidate(inputData2).errors
```
benjaminparker:
```suggestion
        Transport.form().fillAndValidate(inputData1).errors should be (empty) 
```

PauloPortugal: @benjaminparker I have addressed all the above comments. Please review.

test/forms/ConsignmentReferencesSpec.scala
```diff
@@ -75,11 +76,11 @@ class ConsignmentReferencesSpec extends BaseSpec {
       val errors3 = ConsignmentReferences.form().fillAndValidate(inputData3).errors
       errors3 mustBe empty
```
MaciejRewera: I would change those 2 lines into one. And if I remember correctly, we mostly use `must be()` instead of `mustBe`.

test/forms/TransportSpec.scala
```diff
@@ -82,12 +82,12 @@ class TransportSpec extends BaseSpec {
 
     "return no error" when {
 
-      "values are correct" in {
+      "values are correct for different country codes" in {
+        val transportPoland = Transport(Sea, "PL")
+        Transport.form().fillAndValidate(transportPoland).errors.length must be(0)
```
MaciejRewera: I think you could change it to 
```suggestion
        Transport.form().fillAndValidate(transportPoland).errors must be(empty)
```

PR #28, opened 2019-07-12
CEDS-1206 Shut a MUCR journey

MaciejRewera:


app/controllers/DisassociateDucrController.scala
```diff
@@ -19,6 +19,7 @@ package controllers
 import config.AppConfig
 import controllers.actions.{AuthAction, JourneyAction}
 import controllers.util.CacheIdGenerator.movementCacheId
+import controllers.util.FlashKeys
```
ed0906: Is this required? seems its not used

MaciejRewera: It is used when submitting.

ed0906: Ah you've changed the package

app/controllers/util/FlashKeys.scala
```diff
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-package controllers
+package controllers.util
```
ed0906: I think we can come up with a better package for this than `util`

Given Flash is all about data storage how about `storage`/`persistance`

app/views/shut_mucr_confirmation.scala.html
```diff
@@ -0,0 +1,40 @@
...
+@this(main_template: views.html.main_template)
+
+@()(implicit appConfig: AppConfig, request: Request[_], flash: Flash, messages: Messages)
+
+@main_template(
+  title = messages("shutMucrConfirmation.title"),
+  appConfig = appConfig
+) {
+
+  <div class="govuk-box-highlight">
```
ed0906: Maybe we need a component for this summary given we're adding it in for a second time, what are your thoughts?

app/controllers/util/FlashKeys.scala
MaciejRewera: DONE

app/views/shut_mucr_confirmation.scala.html
MaciejRewera: This is good idea. I'll add another commit with it.

PR #29, opened 2019-07-23
CEDS-1204, -1206, -1059 consolidations submissions

MaciejRewera:


app/controllers/ShutMucrController.scala
```diff
@@ -48,10 +53,18 @@ class ShutMucrController @Inject()(
...
+          submissionService.submitShutMucrRequest(shutMucr).map {
+            case ACCEPTED =>
+              Redirect(controllers.routes.ShutMucrConfirmationController.displayPage())
+                .flashing(Flash(Map(FlashKeys.MUCR -> shutMucr.mucr)))
+            case _ => handleError("Unable to submit Shut a Mucr Consolidation request")
         }
       )
   }
+
+  private def handleError(logMessage: String)(implicit request: Request[_]): Result = {
```
PatrykRudnicki: I think we can move this method to `ErrorHandler` it's exactly the same in few controllers

test/base/ExportsTestData.scala
```diff
@@ -1,174 +0,0 @@
...
-import org.joda.time.DateTimeZone.UTC
-import org.joda.time.{DateTime, LocalDate}
-import play.api.libs.json._
-import uk.gov.hmrc.auth.core.AffinityGroup.Individual
-import uk.gov.hmrc.auth.core.retrieve._
-import uk.gov.hmrc.auth.core.{Enrolment, Enrolments, User}
-import uk.gov.hmrc.wco.dec.inventorylinking.common.{AgentDetails, TransportDetails, UcrBlock}
-import uk.gov.hmrc.wco.dec.inventorylinking.movement.request.InventoryLinkingMovementRequest
-
-object ExportsTestData {
```
PatrykRudnicki: ❤️ 😍 😻 💟

test/base/MockFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import scala.concurrent.Future
+
+object MockFactory extends MockitoSugar {
+
+  private val emptyCacheMap = CacheMap("", Map.empty)
+
+  def buildCustomsCacheServiceMock: CustomsCacheService = {
+    val customsCacheServiceMock = mock[CustomsCacheService]
+    when(customsCacheServiceMock.fetchAndGetEntry(any(), any())(any(), any(), any()))
```
PatrykRudnicki: Just my propose about formatting. One empty line after mock and after all whens.

It's difficult to read those code at the first time. I had problem in movements backend.

test/base/MockFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+      .thenReturn(Future.successful(None))
+    when(customsCacheServiceMock.fetch(any())(any(), any())).thenReturn(Future.successful(None))
+    when(customsCacheServiceMock.cache(any(), any(), any())(any(), any(), any()))
+      .thenReturn(Future.successful(emptyCacheMap))
+    when(customsCacheServiceMock.remove(any())(any(), any())).thenReturn(Future.successful(HttpResponse(NO_CONTENT)))
```
PatrykRudnicki: And second comment. I think it's good to create mock definition for every case. Not some general object with everything inside. We can have object with all mocks etc, but this `when` etc. I think it's easier to have it inside specific tests. It's really difficult to reuse those things in other tests, because e.g. you need to have different behaviour or something like this and in practice you need to override this. In more general approach, combination of mocks etc is specific for test case. Thoughts after working with backend movements, it take some time to understand where this mock is, how it works and what mocks. Those stuff are used only in specs, so I think it's enough to create it for specific test cases and don't worry about duplications.

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -16,66 +16,131 @@
...
-import uk.gov.hmrc.wco.dec.MetaData
+import uk.gov.hmrc.play.bootstrap.http.HttpClient
+import uk.gov.hmrc.wco.dec.inventorylinking.movement.request.InventoryLinkingMovementRequest
+import utils.CustomsHeaderNames
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class CustomsDeclareExportsMovementsConnectorSpec
+    extends WordSpec with MustMatchers with MockitoSugar with ScalaFutures {
```
PatrykRudnicki: 👍

test/controllers/AssociateDucrSummaryControllerSpec.scala
```diff
@@ -16,86 +16,182 @@
...
-class AssociateDucrSummaryControllerSpec extends MovementBaseSpec with ViewValidator with BeforeAndAfterEach {
+class AssociateDucrSummaryControllerSpec
+    extends WordSpec with GuiceOneAppPerSuite with MockAuthConnector with ScalaFutures with MustMatchers
+    with URIHelper {
 
   private val uri = uriWithContextPath("/associate-ducr-summary")
 
-  override def beforeEach() {
+  private val submissionServiceMock = buildSubmissionServiceMock
+  private val customsCacheServiceMock = buildCustomsCacheServiceMock
```
PatrykRudnicki: Inlining mock is better. Specially for this case. You receive some methods from this but still below you need to specify different behaviour. Comment related with stuff above.

test/controllers/AssociateDucrSummaryControllerSpec.scala
```diff
@@ -16,86 +16,182 @@
...
+
+      assertThrows[IncompleteApplication] { await(routeGet()) }
     }
 
-    "return Ok for GET request" in {
-      val result = route(app, getRequest(uri)).get
-      status(result) must be(OK)
+    "return Ok for GET request" in new Test {
+
+      status(routeGet()) must be(OK)
```
PatrykRudnicki: Previous stuff slightly easier to read. This shorter. Doesn't matter what will be the result, just want to point this out :)

test/controllers/ChoiceControllerSpec.scala
```diff
@@ -142,10 +142,9 @@ class ChoiceControllerSpec extends MovementBaseSpec with ViewValidator with Befo
       val correctForm =
         JsObject(Map("choice" -> JsString(AllowedChoiceValues.ShutMucr)))
       val result = route(app, postRequest(choiceUri, correctForm)).get
-      val header = result.futureValue.header
 
       status(result) must be(SEE_OTHER)
-      header.headers.get("Location") must be(Some("/customs-movements/shut-mucr"))
+      redirectLocation(result) must be(Some(routes.ShutMucrController.displayPage().url))
```
PatrykRudnicki: :+1

test/controllers/DisassociateDucrControllerSpec.scala
```diff
@@ -16,72 +16,140 @@
...
+  import DisassociateDucrControllerSpec._
 
-class DisassociateDucrControllerSpec extends MovementBaseSpec with BeforeAndAfterEach {
+  implicit val defaultPatience: PatienceConfig =
+    PatienceConfig(timeout = Span(5, Seconds), interval = Span(500, Millis))
 
   private val uri = uriWithContextPath("/disassociate-ducr")
 
-  trait SetUp {
+  private val submissionServiceMock = buildSubmissionServiceMock
```
PatrykRudnicki: The same with mock here, below you need to add another scenarios. Slightly messy

test/controllers/DisassociateDucrControllerSpec.scala
```diff
@@ -16,72 +16,140 @@
...
+          .thenReturn(Future.successful(BAD_REQUEST))
+
+        status(routePost()) must be(INTERNAL_SERVER_ERROR)
+      }
+    }
+
+    def routePost(headers: Map[String, String] = Map.empty, body: JsValue = correctForm): Future[Result] =
+      route(
+        app,
+        FakeRequest(POST, uri)
```
PatrykRudnicki: Whole this method can go to some helper trait. It's used in few places (or will be used)

test/controllers/ShutMucrControllerSpec.scala
```diff
@@ -43,11 +45,12 @@ import utils.FakeRequestCSRFSupport._
...
 class ShutMucrControllerSpec
-    extends WordSpec with GuiceOneAppPerSuite with MockAuthConnector with ScalaFutures with MustMatchers {
+    extends WordSpec with GuiceOneAppPerSuite with MockAuthConnector with ScalaFutures with MustMatchers
+    with URIHelper {
 
-  private val shutMucrUri = "/customs-movements/shut-mucr"
+  private val shutMucrUri = uriWithContextPath("/shut-mucr")
 
-  private val submissionServiceMock = mock[SubmissionService]
+  private val submissionServiceMock = buildSubmissionServiceMock
```
PatrykRudnicki: The same here

test/controllers/ShutMucrControllerSpec.scala
```diff
@@ -105,17 +113,31 @@ class ShutMucrControllerSpec
...
 
-        authorizedUser()
         val result = routePost(body = incorrectShutMucrJSON)
 
         contentAsString(result) must include(messages("shutMucr.title"))
       }
+
+      "not call SubmissionService" in new Test {
+
+        routePost(body = incorrectShutMucrJSON)
```
PatrykRudnicki: For every case like here you should use `futureValue`. Line below can be invoked without waiting until this is finished

test/resources/inventory-linking-exports-schemas/common/inventoryLinkingCommonTypes.xsd
```diff
@@ -0,0 +1,249 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
PatrykRudnicki: Do we need those schemas here?

app/controllers/ShutMucrController.scala
MaciejRewera: It depends. If we mvoe it to `ErrorHandler` then we would need to pass logger as an argument. Having log errors all being called by one object is hard to track.
I know that it breaks DRY rules, but the only purpose of this method is actually to log the error. I'll leave it for now.

test/base/MockFactory.scala
MaciejRewera: 👍

test/base/MockFactory.scala
MaciejRewera: I totally agree with you. The reason behind creating this `MockFactory` is not to provide mocks ready to use in every test case. It is there because Mockito does not work perfectly with Scala (so far) and in most cases the default return value is `null`. To avoid `null`s I decided to create `MockFactory`, to define default values that match return types.

test/controllers/ShutMucrControllerSpec.scala
MaciejRewera: Fixed

test/resources/inventory-linking-exports-schemas/common/inventoryLinkingCommonTypes.xsd
MaciejRewera: Pasted them to be easy to reach. Could think about automatically updating if ILE changes them.

test/controllers/DisassociateDucrControllerSpec.scala
MaciejRewera: Created a class to handle routing GET and POST. Thanks for the idea :+1:

PR #30, opened 2019-07-24
CEDS-1072 Show Loaders ILE Notifications

PauloPortugal:


test/controllers/ChoiceControllerSpec.scala
```diff
@@ -26,7 +26,7 @@ import org.scalatest.BeforeAndAfterEach
 import play.api.libs.json.{JsObject, JsString}
 import play.api.test.Helpers._
 
-class ChoiceControllerSpec extends MovementBaseSpec with ViewValidator with BeforeAndAfterEach {
+class ChoiceControlleqrSpec extends MovementBaseSpec with ViewValidator with BeforeAndAfterEach {
```
PatrykRudnicki: Please remember about removing `q` :)

app/models/Notification.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package models
+import java.time.LocalDateTime
+
+import play.api.libs.json.Json
+
+case class Notification(dateTimeReceived: LocalDateTime, conversationId: String) extends Ordered[Notification] {
+
+  def compare(that: Notification): Int =
+    if (this.dateTimeReceived == that.dateTimeReceived) 0
```
ed0906: Could just subtract the two dates?

app/models/Notification.scala
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+import java.time.LocalDateTime
+
+import play.api.libs.json.Json
+
+case class Notification(dateTimeReceived: LocalDateTime, conversationId: String) extends Ordered[Notification] {
```
ed0906: How come we're using `LocalDateTime`?

If we run this service in two timezones we would get two different LocalDateTimes for what is essentially the same time. It doesn't carry any timezone info it would falsely appear in the DB as being different.

We should use `Instant` which is a UTC date

app/models/Submission.scala
```diff
@@ -0,0 +1,33 @@
...
+import java.time.LocalDateTime
+
+import play.api.libs.json.Json
+
+case class Submission(
+  conversationId: String,
+  ucr: String,
+  submissionType: String,
+  submissionAction: String,
+  dateUpdated: LocalDateTime,
```
ed0906: Similar to above, the recommended approach with Dates is to use `Instant`

test/views/NotificationsViewSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+class NotificationsViewSpec extends WordSpec with MustMatchers with Stubs with ViewValidator {
+
+  val messages = stubMessages()
+  val page: Html = new notifications(mainTemplate)(Seq.empty)(FakeRequest(), minimalAppConfig, messages)
+
+  "Notification page" should {
+
+    "contains title" in {
+
+      getElementById(page, "title").text() must be(messages("notifications.title"))
```
ed0906: Theres some nice matchers you can use for these e.g.

`page.getElementById("title") should containText(messages("notifications.title"))`

app/models/Notification.scala
PatrykRudnicki: `compare` is just to be able to sort dates. I's not related with substracting etc.

app/models/Notification.scala
PatrykRudnicki: Here is the same approach like in exports frontend. We just copied this.

But the question is good. I think we can change LocalDateTime, but we need to do it in every microservice. Not only here. I feel new ticket for it.

app/models/Notification.scala
ed0906: Every time you do an `if` you need two tests, so to work out if one date is bigger than another this method essentially needs 3 unit tests.

In fact all you have to do is `this.dateTimeReceived.compareTo(that.dateTimeReceived)`
If you dig into the implementation of that method though all it does is subtract one date from the other

app/models/Notification.scala
PauloPortugal: @ed0906 `LocalDateTime` does not carry any timezone, so that would not be an issue.
`ZonedDateTime` keeps a timezone and in that instance we would see a different time if we kept the value in the DB as UTC (zulu time).

Am I correct?

app/models/Submission.scala
PauloPortugal: `LocalDateTime` vs `ZonedDateTime`
I am getting confused now. Please confirm :)

app/models/Notification.scala
PauloPortugal: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDateTime.html
`
This class does not store or represent a time-zone. Instead, it is a description of the date, as used for birthdays, combined with the local time as seen on a wall clock. It cannot represent an instant on the time-line without additional information such as an offset or time-zone.
`

PauloPortugal: So, as far as I understand, as long as we capture the client's local time correctly and store it in the DB and present it back it is fine _**as long as the client does not change timezone**_ (not the server).
If submit a declaration in the UK today at 13h and then view it in Germany the following day I should see today's date and 14h (if we kept the zone information _Instant/UTC/ZonedDateTime_), but If I saw the time 13h it would be OKish.
We should use ZonedDateTime and store the date/time in an epock/instant format in the DB.
But I think this is a user need requirement, although heavily influenced by the technological implementation, but this requires a further discussion with BAs to understand what should a customer see on our FE? A simple LocalDateTime or a ZonedDateTime?

app/models/Notification.scala
PauloPortugal: This has been addressed.

test/views/NotificationsViewSpec.scala
PauloPortugal: This has been addressed.

app/models/Notification.scala
ed0906: @PauloPortugal It's exactly because LocalDateTime doesn't contain timezone info that *is* the issue. We either want to use UTC time (`Instant`) or some date type *with* a timezone (`ZonedDateTime`). Given `Instant` is generally easier to use given the average operation we tend to do (e.g. saving it in a DB or converting to a String) I think the general consensus is to use that

ed0906: > @ed0906 `LocalDateTime` does not carry any timezone, so that would not be an issue.
> `ZonedDateTime` keeps a timezone and in that instance we would see a different time if we kept the value in the DB as UTC (zulu time).
> 
> Am I correct?

The JSON serialization for Mongo would convert it to a Mongo Date Type which is basically a mirror image of what  an `Instant` is (number of milliseconds since 1/1/1970 in UTC time). As a result ZonedDateTime would be fine as the serialization process would know how to get UTC time from it. 
With LocalDateTimes how do you convert that to UTC? You have to *presume* the time zone which is why we shouldnt use them.

app/models/Submission.scala
ed0906: ZonedDateTime is better than LocalDateTime, but Instant makes everyones life easier

ed0906: For the avoidance of any doubt I suggest we *always* use `Instant` :)

app/models/Notification.scala
PatrykRudnicki: Guys, as I understand LocalDateTime will be date from the server, not specific user, so we don't have problem with this. It's not necessary to capture zone etc.

PatrykRudnicki: From documentation: `Obtains the current date-time from the system clock in the default time-zone.`

PauloPortugal: I have changed to ZonedDateTime as it carries information about the timezone.
@MaciejRewera please adjust the DB to store an instant/epoch

test/views/NotificationsViewSpec.scala
ed0906: Apologies I wasn't saying what you'd done! It was just for future reference

PR #31, opened 2019-07-30
CEDS-1072 Rename the submissions to movements

PauloPortugal:
We do not really use that domain language here and we should not bring
submissions to the mix. These are movements.
This also included a fix for the movements backend url.

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -86,7 +86,7 @@ microservice {
       save-movement-uri = "/save-movement-submission"
       submit-consolidation = "/consolidations/submit"
       fetch-notifications = "/notifications"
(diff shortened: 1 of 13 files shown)


conf/application.conf
```diff
@@ -86,7 +86,7 @@ microservice {
       save-movement-uri = "/save-movement-submission"
       submit-consolidation = "/consolidations/submit"
       fetch-notifications = "/notifications"
-      fetch-submissions = "/submissions"
+      fetch-movements = "/movements"
```
PatrykRudnicki: You need to change endpoint in movements backend

PauloPortugal: `movements` is the correct one :)

PR #39, opened 2019-07-31
CEDS-1436 Replace Notification model with NotificationPresentationData

MaciejRewera:
Since this class is used for presentation only, it contains only
presentation-relevant fields. Backend should send Notification data
in the same format, decoupling presentation model from DB model.

### DIFF
--- a/app/views/movements.scala.html
+++ b/app/views/movements.scala.html
@@ -18,11 +18,11 @@
 @import java.time.format.DateTimeFormatter
 
 @import config.AppConfig
-@import models.{Notification, Movement}
+@import models.{Movement, NotificationPresentation}
 
 @this(main_template: views.html.main_template)
 
-@(submissions: Seq[(Movement, Seq[Notification])])(implicit request: Request[_], appConfig: AppConfig, messages: Messages)
+@(submissions: Seq[(Movement, Seq[NotificationPresentation])])(implicit request: Request[_], appConfig: AppConfig, messages: Messages)
 
 @main_template(
     title = messages("submissions.title"),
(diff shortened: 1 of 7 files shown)


app/models/NotificationPresentationData.scala
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package models
+
+import java.time.Instant
+
+import play.api.libs.json.Json
+
+final case class NotificationPresentationData(
```
ed0906: Any chance we can remove the `Data` suffix? it doesn't add much value

app/views/movements.scala.html
```diff
@@ -49,7 +49,7 @@
                     <td class="table-cell" id="submissionType-@submission.conversationId">@submission.submissionType</td>
                     <td class="table-cell" id="submissionAction-@submission.conversationId">@submission.submissionAction</td>
                     <td class="table-cell" id="dateUpdated-@submission.conversationId">
-                        @{notifications.headOption.map(n => n.dateTimeReceived).map(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm").withZone(ZoneId.systemDefault()).format(_))}</td>
+                        @{notifications.headOption.map(n => n.timestampReceived).map(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm").withZone(ZoneId.systemDefault()).format(_))}</td>
```
SteveSugden: (Not part of this PR change, but...) Would it not be useful to extract this date formatting logic somewhere so that it can be reused by other parts of the UI?

ed0906: Yeh Ideally we should pull it into a class which we can unit test and share across modules

ed0906: Also `n => n.timestampReceived` is the same as `_.timestampReceived`

app/models/NotificationPresentationData.scala
MaciejRewera: Done

PR #40, opened 2019-08-01
CEDS Update Local Auth Redirect

ed0906:


app/handlers/ErrorHandler.scala
```diff
@@ -43,7 +43,7 @@ class ErrorHandler @Inject()(appConfig: AppConfig, val messagesApi: MessagesApi,
 
   override def resolveError(rh: RequestHeader, ex: Throwable): Result =
     ex match {
-      case _: NoActiveSession => toGGLogin(rh.uri)
```
ed0906: `toGGLogin` only works for production as far as I'm aware. You can override the host but that isn't sufficient as we need to redirect to the auth-stub

PR #43, opened 2019-08-06
CEDS-1418 Requests Distinction

MaciejRewera:
This PR is introduced to allow differentiating between different submission types being made by the user.
To fulfill that, every request has its own endpoint on the backend, since this is the simplest solution. It involves almost no logic and is more clean than in-code-logic solution.


app/views/movements.scala.html
```diff
@@ -36,18 +36,16 @@
...
-                <th class="table-cell" id="status">@messages("submissions.status")</th>
                 <th class="table-cell" id="noOfNotifications">@messages("submissions.noOfNotifications")</th>
             </tr>
 
             @for( (submission, notifications) <- submissions ){
                 <tr class="table-row">
-                    <td class="table-cell" id="ucr-@submission.conversationId">@submission.ucr</td>
-                    <td class="table-cell" id="submissionType-@submission.conversationId">@submission.submissionType</td>
-                    <td class="table-cell" id="submissionAction-@submission.conversationId">@submission.submissionAction</td>
+                    <td class="table-cell" id="ucr-@submission.conversationId">@submission.ucrBlocks.headOption.map(_.ucr).getOrElse("-")</td>
```
bambuchaAdm: And what with other ucr's blocks ?

MaciejRewera: Thanks for spotting this. Don't know if that will ever be the case, but added fix for that.

app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -33,47 +32,66 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
 
   private val logger = Logger(this.getClass)
 
-  private val MovementSubmissionUrl = s"${appConfig.customsDeclareExportsMovements}${appConfig.saveMovementSubmission}"
-  private val MovementConsolidationUrl =
-    s"${appConfig.customsDeclareExportsMovements}${appConfig.submitMovementConsolidation}"
+  private val CustomsDeclareExportsMovementsUrl = s"${appConfig.customsDeclareExportsMovements}"
+  private val ArrivalSubmissionUrl =
```
bambuchaAdm: why `val`s from capital letter ?

MaciejRewera: To indicate they're constants holding specific value.

PR #47, opened 2019-08-09
CEDS-1485 Add test for displaying all UCRs on Movements page

MaciejRewera:


app/models/SubmissionPresentation.scala
```diff
@@ -32,4 +32,12 @@ case class SubmissionPresentation(
 
 object SubmissionPresentation {
   implicit val formats = Json.format[SubmissionPresentation]
+
+  object ActionTypes {
+    val Arrival = "Arrival"
```
bambuchaAdm: Why strings? Maybe case objects with common `sealed abstract class`

test/views/MovementsViewSpec.scala
```diff
@@ -89,5 +91,30 @@ class MovementsViewSpec extends WordSpec with MustMatchers with Stubs with ViewV
       getElementById(pageWithData, s"dateOfUpdate-$conversationId").text() must be("2019-10-31 00:10")
       getElementById(pageWithData, s"noOfNotifications-$conversationId").text() must be("1")
     }
+
+    "contain MUCR and DUCR if Submission contains both" in {
+      val pageWithData: Html = new movements(mainTemplate)(
+        Seq(
+          (
```
bambuchaAdm: Is this is for tuple ? Maybe `->` would looks better?

MaciejRewera: DONE

app/models/SubmissionPresentation.scala
MaciejRewera: Yea, good idea. Changing that.

bambuchaAdm:
:+1: for test on writes

PR #52, opened 2019-08-23
CEDS-1518: Improve ILE Requests Screen

mattclark-zerogravit:


app/views/components/fields/FieldContentFormatter.scala
```diff
@@ -21,8 +21,8 @@ import java.time.{Instant, ZoneId}
 
 object FieldContentFormatter {
 
-  private val dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm").withZone(ZoneId.systemDefault())
+  private val dateTimeFormatter = DateTimeFormatter.ofPattern("dd MMM yyyy - HH:mm").withZone(ZoneId.systemDefault())
```
bambuchaAdm: you can put 'at' instead - and replace

mattclark-zerogravit: doesn;t work. its not a valid pattern

bambuchaAdm: "dd MMM yyyy 'at' HH:mm" - something like this

PR #55, opened 2019-08-29
CEDS-1552 Reformat codebase and removed unused imports

PatrykRudnicki:


app/models/viewmodels/Decoder.scala
```diff
@@ -27,11 +27,9 @@ class Decoder {
 
   def soe(code: String): String = soeMappings(code)
 
-  private val crcMappings: Map[String, String] = Map(
-    "000" -> "Success",
-    "101" -> "Pre-lodged Declaration has not arrived ",
-    "102" -> "Declaration has not arrived"
-  ).withDefaultValue("")
+  private val crcMappings: Map[String, String] =
+    Map("000" -> "Success", "101" -> "Pre-lodged Declaration has not arrived ", "102" -> "Declaration has not arrived")
```
bambuchaAdm: @PatrykRudnicki huh... so ugly...

bambuchaAdm:
Please enable task failing on `scalafmt::test`

app/models/viewmodels/Decoder.scala
PatrykRudnicki: ScalaFormatter, I added failing when code is not formatted

PR #60, opened 2019-09-02
CEDS-1587 Use descriptive name for metrics

bambuchaAdm:
Now are too domain specific and does not carry any information


bambuchaAdm:
test this please

PR #61, opened 2019-09-02
CEDS-1509 enhancements

MaciejRewera:
This PR introduces some changes mainly to Decoder and NotificationPageSingleElementFactory classes.
The biggest difference here are abstract `Code` classes that are supposed to ensure some type-safety.


MaciejRewera:
Force pushed because I did not want to create a separate commit for one line formatter change.

PR #65, opened 2019-09-04
Ceds 1602 - Review logging level

bambuchaAdm:
Adjust logging level to platform guides.

Most of non application errors (missing data in db, bad requests) should be on warn level rather then error.

### DIFF
--- a/app/connectors/CustomsDeclareExportsMovementsConnector.scala
+++ b/app/connectors/CustomsDeclareExportsMovementsConnector.scala
@@ -27,6 +27,7 @@ import uk.gov.hmrc.http.{HeaderCarrier, HttpResponse}
 import uk.gov.hmrc.play.bootstrap.http.HttpClient
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
 
 @Singleton
 class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, httpClient: HttpClient) {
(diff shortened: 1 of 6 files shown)


app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -51,59 +52,66 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
   def sendArrivalDeclaration(
     requestXml: String
   )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] =
-    postRequest(ArrivalSubmissionUrl, requestXml)
+    postRequest(ArrivalSubmissionUrl, requestXml, "arrival")
```
SteveSugden: Don't we have constants for these journeys/actions somewhere?

app/controllers/consolidations/AssociateDucrSummaryController.scala
```diff
@@ -71,13 +73,16 @@ class AssociateDucrSummaryController @Inject()(
         case ACCEPTED =>
           Redirect(routes.AssociateDucrConfirmationController.displayPage())
             .flashing(FlashKeys.MUCR -> mucrOptions.mucr)
-        case _ => handleError("Unable to submit Association Consolidation request")
+        case _ =>
+          MDC.put("MUCR", mucrOptions.mucr)
```
SteveSugden: What's this doing?

SteveSugden: I've seen/used MDC used to add a variable into log messages before.  But isn't this normally done at the beginning of the request handler?  Not just before you log.

bambuchaAdm: There is default filter which would add all information from `HeaderCarrier` to `MDC`. But here we would like to have additional information about MUCR to be better searchable in kibana.

SteveSugden: So what does the output of that warning log line below look like?  You are already adding the MUCR and DUCR to the log message.  I'll have to come and have a chat tomorrow.

app/controllers/SummaryController.scala
```diff
@@ -53,9 +53,13 @@ class SummaryController @Inject()(
...
-      case Some(data) if typeOfJourney == Arrival   => Ok(arrivalSummaryPage(data))
-      case Some(data) if typeOfJourney == Departure => Ok(departureSummaryPage(data))
-      case _                                        => handleError("Could not obtain data from DB")
+      case Some(data) if typeOfJourney == Arrival =>
+        Ok(arrivalSummaryPage(data))
+      case Some(data) if typeOfJourney == Departure =>
+        Ok(departureSummaryPage(data))
+      case _ =>
+        logger.warn(s"No movement data found in cache.")
+        errorHandler.getInternalServerErrorPage
```
PatrykRudnicki: Prefer previous version, but of course with changed method and using `errorHandler.getInternalServerErrorPage` instead of 
```
InternalServerError(
      errorHandler.standardErrorTemplate(
        pageTitle = Messages("global.error.title"),
        heading = Messages("global.error.heading"),
        message = Messages("global.error.message")
      )
    )
```

So in general I mean it's better to have it in small method

app/controllers/consolidations/AssociateDucrSummaryController.scala
PatrykRudnicki: Yeah, I want to see log too

PatrykRudnicki: Is it will add just a column on the kibana to have possibility to see/filter/find by MUCR/DUCR?

bambuchaAdm: I kibna
* message: Unable to submit Association Consolidation request: MUCR XXXX and DUCR YYY
* ducr: YYY
* mucr: XXXX

SteveSugden: I think I get it now.   Kibana keys are the result of keys in MDC?   Seems a bit clunky, but if that's what's required.

app/controllers/SummaryController.scala
bambuchaAdm: Maybe it should be move to local action Refiner ?

PR #66, opened 2019-09-05
BAU - TamperMonkey scripts for Arrival and Departure journey pages

SteveSugden:


bambuchaAdm:
From perspective of usability - would you like to add `@updateURL` to github RAW url and `@version` to script

PR #74, opened 2019-09-09
CEDS-1529 Add ArrivalReference to InventoryLinkingExportsRequest

PatrykRudnicki:


app/forms/Movement.scala
```diff
@@ -46,14 +46,17 @@ object Movement {
 
     val transport = cacheMap.getEntry[Transport](Transport.formId)
 
+    val arrivalReference = cacheMap.getEntry[ArrivalReference](ArrivalReference.formId).flatMap(_.reference)
```
MaciejRewera: Is there a reason why it is only arrival? What about departure? As far as I remember, the `movementReference` is something we need to have for both.

PatrykRudnicki: The page is created only for arrival. As I know from BA and designer it's only for arrival journey.

MaciejRewera: :+1:

PR #77, opened 2019-09-10
CEDS-1618 Second round of moving tests to be unit

PatrykRudnicki:


test/unit/base/ControllerSpec.scala
```diff
@@ -43,4 +44,8 @@ trait ControllerSpec
     FakeRequest("POST", "")
       .withJsonBody(body)
       .withCSRFToken
+
+  protected def mockArrivalJourney(): Unit = withCaching(Choice.choiceId, Some(Choice(AllowedChoiceValues.Arrival)))
```
ed0906: I think it would read better in unit tests if this was called something like:
"givenIAmOnTheArrivalJourney"
"givenAUserOnTheArrivalJourney"
etc

When I read the unit tests and I see "mockArrivalJourney" my first thought at the moment is "what does that do?"

test/unit/controllers/LocationControllerSpec.scala
```diff
@@ -0,0 +1,146 @@
...
+
+        status(result) mustBe OK
+        theResponseForm.value.value mustBe cachedData
+      }
+
+      "return Error" when {
+
+        "no JourneyType found" in {
+
+          mockArrivalJourney()
```
ed0906: Looks like this mock gets overridden immediately after

test/unit/base/ControllerSpec.scala
PatrykRudnicki: Okay, `givenAUserOnTheArrivalJourney` sounds better for me

ed0906: Similar for the method below

test/unit/controllers/LocationControllerSpec.scala
PatrykRudnicki: Fixed

test/unit/base/ControllerSpec.scala
PatrykRudnicki: Fixed

PR #84, opened 2019-09-13
CEDS-1637 error messages

MaciejRewera:


app/models/notifications/NotificationFrontendModel.scala
```diff
@@ -28,8 +28,9 @@ final case class NotificationFrontendModel(
   ucrBlocks: Seq[UcrBlock],
   masterRoe: Option[String],
   masterSoe: Option[String],
+  crcCode: Option[String],
```
PatrykRudnicki: You changed the order of arguments. Can you confirm this? Both of the arguments has the same type `Option[String]` so in places where we can use it it'll contains wrong data.

app/models/viewmodels/NotificationPageSingleElementFactory.scala
```diff
@@ -84,14 +83,30 @@ class NotificationPageSingleElementFactory @Inject()(decoder: Decoder) {
...
+    } yield content
+
+    val errorExplanation = (for {
+      code <- notification.errorCodes
+      errorCode <- decoder.errorCode(code) match {
+        case None =>
+          logger.warn(s"Received inventoryLinkingControlResponse with unknown error code: $code")
+          None
+        case knownCode => knownCode
+      }
```
PatrykRudnicki: This code return the same element. What do you think about something like this

```
def method(a: Option[String]): Option[String] =
  if(a.isEmpty) logger.warn(...)
  a
}
```

Of course with better naming or without function

```
errorCode <- {
  val code = decoder.errorCode(code)
  if(code.isEmpty) logger.warn(...)
  code
}
```

test/models/viewmodels/decoder/RoeCodeSpec.scala
```diff
@@ -41,7 +40,7 @@ class RoeCodeSpec extends UnitSpec {
         PrelodgePrefix
       )
 
-      RoeCode.codes mustBe expectedCodes
+      RoeCode.codes mustEqual expectedCodes
```
PatrykRudnicki: Why `mustEqual`?

app/models/viewmodels/NotificationPageSingleElementFactory.scala
MaciejRewera: :+1:

app/models/notifications/NotificationFrontendModel.scala
MaciejRewera: Double checked, run tests. Everything looks fine and working.

app/models/viewmodels/NotificationPageSingleElementFactory.scala
MaciejRewera: Used you second solution.

test/models/viewmodels/decoder/RoeCodeSpec.scala
MaciejRewera: Changed back to `mustBe` in a separate commit. In this case there is no difference in functionality, but should compile slightly faster.

app/models/viewmodels/NotificationPageSingleElementFactory.scala
```diff
@@ -84,14 +83,30 @@ class NotificationPageSingleElementFactory @Inject()(decoder: Decoder) {
     notification: NotificationFrontendModel
   )(implicit messages: Messages): NotificationsPageSingleElement = {
 
-    val content = notification.actionCode.map { code =>
-      s"<p>${messages(decoder.actionCode(code).map(_.contentKey).getOrElse(""))}</p>"
-    }
+    val actionCodeExplanation = for {
+      code <- notification.actionCode
+      actionCode <- decoder.actionCode(code)
+      content = s"<p>${messages(actionCode.contentKey)}</p>"
```
bambuchaAdm: why in loop

app/models/viewmodels/decoder/ErrorCode.scala
```diff
@@ -0,0 +1,210 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.viewmodels.decoder
+
+sealed abstract class ErrorCode(override val code: String, override val status: String, override val contentKey: String)
```
bambuchaAdm: scaladoc

PR #87, opened 2019-09-18
CEDS-1646 Move Request Notification outside of Notifications List

ed0906:


app/controllers/NotificationsController.scala
```diff
@@ -38,18 +39,22 @@ class NotificationsController @Inject()(
...
-
-      notificationsElems <- connector
-        .fetchNotifications(conversationId)
-        .map(notifications => notifications.sorted.map(factory.build))
-
-      submissionUcr = submission.flatMap(extractUcr).getOrElse("")
-      elementsToDisplay = (submissionElem.toSeq ++ notificationsElems)
-
-    } yield Ok(notifications(submissionUcr, elementsToDisplay))
+    val params = for {
```
PatrykRudnicki: I think with map it'll be easier to read and shorter, but it's just my opinion, non blocking comment

ed0906: I tried various methods, using a map, using chained for comprehension. None seemed to compile :(

PR #88, opened 2019-09-18
Notifications page element factory refactor

MaciejRewera:
This PR sets ground for the tickets coming next for notifications page.
In one of the commits it addressed the change of logging level for unknown notification error codes. For more details look into commit message.

### DIFF
--- a/app/models/viewmodels/notificationspage/ControlResponseConverter.scala
+++ b/app/models/viewmodels/notificationspage/ControlResponseConverter.scala
@@ -52,7 +52,7 @@ private[notificationspage] class ControlResponseConverter @Inject()(decoder: Dec
         code <- notification.errorCodes
         errorCode <- {
           val errorCode = decoder.errorCode(code)
-          if (errorCode.isEmpty) logger.warn(s"Received inventoryLinkingControlResponse with unknown error code: $code")
+          if (errorCode.isEmpty) logger.info(s"Received inventoryLinkingControlResponse with unknown error code: $code")
 
           errorCode
         }
--- a/app/models/viewmodels/notificationspage/ControlResponseConverter.scala
+++ b/app/models/viewmodels/notificationspage/ControlResponseConverter.scala
@@ -45,7 +45,7 @@ private[notificationspage] class ControlResponseConverter @Inject()(decoder: Dec
       val actionCodeExplanation = for {
         code <- notification.actionCode
         actionCode <- decoder.actionCode(code)
-        content = s"<p>${messages(actionCode.contentKey)}</p>"
+        content = paragraph(messages(actionCode.contentKey))
       } yield content
 
       val errorExplanation = (for {
@@ -57,7 +57,7 @@ private[notificationspage] class ControlResponseConverter @Inject()(decoder: Dec
           errorCode
         }
 
-        content = s"<p>${messages(errorCode.contentKey)}</p>"
+        content = paragraph(messages(errorCode.contentKey))
       } yield content).foldLeft("")(_ + _)
 
       NotificationsPageSingleElement(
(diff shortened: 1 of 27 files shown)


app/models/viewmodels/notificationspage/ControlResponseConverter.scala
```diff
@@ -0,0 +1,77 @@
...
+import javax.inject.{Inject, Singleton}
+import models.notifications.NotificationFrontendModel
+import models.notifications.ResponseType.ControlResponse
+import models.viewmodels.decoder.Decoder
+import play.api.Logger
+import play.api.i18n.Messages
+import play.twirl.api.Html
+
+@Singleton
+private[notificationspage] class ControlResponseConverter @Inject()(decoder: Decoder)
```
bambuchaAdm: why package private ?

app/models/viewmodels/notificationspage/MovementResponseConverter.scala
```diff
@@ -0,0 +1,71 @@
...
+    notification.responseType == MovementResponse
+
+  override def convert(
+    notification: NotificationFrontendModel
+  )(implicit messages: Messages): NotificationsPageSingleElement =
+    if (canConvertFrom(notification)) {
+
+      val crcCodeContent = getContentForCrcCode(notification)
+      val content = crcCodeContent.map { content =>
+        paragraph(s"${messages("notifications.elem.content.inventoryLinkingMovementResponse.crc")} $content")
```
bambuchaAdm: `messages("notifications.elem.content.inventoryLinkingMovementResponse.crc")` would be nicer as shorter variable defined earlier

app/models/viewmodels/notificationspage/ControlResponseConverter.scala
MaciejRewera: Because I don't expect it to be used outside. These are actually private objects to the Factory in some sense. Their content is the Factory's implementation which would be private otherwise.

app/models/viewmodels/notificationspage/MovementResponseConverter.scala
MaciejRewera: Do you mean just to extract this long line into a variable? Similar to `paragraph` ?

test/models/viewmodels/notificationspage/ControlResponseConverterSpec.scala
```diff
@@ -0,0 +1,220 @@
...
+class ControlResponseConverterSpec extends BaseSpec with MockitoSugar {
+
+  private val testTimestampString = "2019-10-23T12:34+00:00"
+  private val formatter = DateTimeFormatter.ISO_OFFSET_DATE_TIME.withZone(ZoneId.systemDefault())
+  private val testTimestamp = ZonedDateTime.parse(testTimestampString, formatter).toInstant
+
+  private val AcknowledgedAndProcessedActionCode = ActionCode.AcknowledgedAndProcessed
+
+  private trait Test {
+    implicit val messages: Messages = Mockito.spy(stubMessages())
```
bambuchaAdm: why spy on messages ? It is unused later

app/models/viewmodels/notificationspage/MovementResponseConverter.scala
```diff
@@ -0,0 +1,71 @@
...
+import models.notifications.ResponseType.MovementResponse
+import models.viewmodels.decoder.Decoder
+import play.api.i18n.Messages
+import play.twirl.api.Html
+
+@Singleton
+private[notificationspage] class MovementResponseConverter @Inject()(decoder: Decoder)
+    extends NotificationPageSingleElementConverter {
+
+  private val dateTimeFormatter = DateTimeFormatter.ofPattern("dd MMM yyyy 'at' HH:mm").withZone(ZoneId.systemDefault())
```
bambuchaAdm: why not inject ?

MaciejRewera: I decided to use constant in companion object.

PR #89, opened 2019-09-18
CEDS-1660 Audit Movements Arrival Submission

PauloPortugal:


app/services/audit/AuditService.scala
```diff
@@ -0,0 +1,93 @@
...
+import play.api.libs.json.{JsObject, Json}
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.audit.AuditExtensions
+import uk.gov.hmrc.play.audit.http.connector.AuditResult.{Disabled, Failure, Success}
+import uk.gov.hmrc.play.audit.http.connector.{AuditConnector, AuditResult}
+import uk.gov.hmrc.play.audit.model.{DataEvent, ExtendedDataEvent}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AuditService @Inject()(connector: AuditConnector, appConfig: AppConfig)(implicit ec: ExecutionContext) {
```
PatrykRudnicki: Instead of injecting whole AppConfig I think you will be able to inject just appName: `@Named("appName") appName: String`

app/config/AppConfig.scala
```diff
@@ -30,7 +30,7 @@ class AppConfig @Inject()(
   val runModeConfiguration: Configuration,
   val environment: Environment,
   servicesConfig: ServicesConfig,
-  @Named("appName") appName: String
+  @Named("appName") val appName: String
```
PatrykRudnicki: in case if you will be able to inject just appName you don't need val here

PauloPortugal: This has been addressed

app/services/audit/AuditService.scala
PauloPortugal: This has been addressed.

PR #90, opened 2019-09-18
CEDS-1646 Convert <P> to <Div>

ed0906:


app/views/notifications.scala.html
```diff
@@ -28,7 +28,7 @@
         <div id="notifications-request">
             <h2 id="notifications-request-title" class="heading-large">@submissionElement.title</h2>
             <p id="notifications-request-timestamp">@{submissionElement.timestampInfo}</p>
-            <p id="notifications-request-content">@{submissionElement.content}</p>
+            <div id="notifications-request-content">@{submissionElement.content}</div>
```
SteveSugden: Why just this <p>?   Why not the timestamp one above?

ed0906: Its because the content is a `Html` element, for some reason it didn't render properly when using paragraph tags

PR #91, opened 2019-09-19
CEDS-1639 backend contract

MaciejRewera:


app/models/viewmodels/notificationspage/MovementTotalsResponseConverter.scala
```diff
@@ -39,8 +39,10 @@ class MovementTotalsResponseConverter @Inject()(decoder: Decoder, dateTimeFormat
     if (canConvertFrom(notification)) {
 
       val crcCodeExplanation = notification.crcCode.flatMap(buildCrcCodeExplanation)
-      val roeCodeExplanation = notification.masterRoe.flatMap(buildRoeCodeExplanation)
-      val soeCodeExplanation = notification.masterSoe.flatMap(buildSoeCodeExplanation)
+      val roeCodeExplanation =
+        notification.entries.headOption.flatMap(_.entryStatus).flatMap(_.roe).flatMap(buildRoeCodeExplanation)
```
SteveSugden: Have I understood correctly?  There can only ever be max one Entry?   In which case, why is it stored as a Seq?

bambuchaAdm: Looks like there is some logic connected to that `Seq` - it would be nice to have `Entries` object which would encapsulate logic related this Seq to separate object.

This is concept called [first class collection](https://softwareengineering.stackexchange.com/questions/139353/why-should-we-preferably-use-first-class-collections) which brings much more

MaciejRewera: According to the schema, there might be several Entries. Even in our case, for association we will have MUCR and DUCR Entries. The reason it is only processing the first Entry is not to change the behaviour. It will be changed in the next PR as part of the ticket.

PR #92, opened 2019-09-19
CEDS-1659 Showing correct message for ducr and mucr during movements

PatrykRudnicki:


app/models/viewmodels/notificationspage/NotificationPageSingleElementFactory.scala
```diff
@@ -48,8 +48,10 @@ class NotificationPageSingleElementFactory @Inject()(
     submission: SubmissionFrontendModel
   )(implicit messages: Messages): NotificationsPageSingleElement = {
 
+    val ucrMessage = if (submission.hasMucr) "MUCR" else "DUCR"
```
MaciejRewera: 👍

PR #96, opened 2019-09-20
Ceds 1638 1639 ers emr messages

MaciejRewera:


app/models/viewmodels/notificationspage/NotificationPageSingleElementFactory.scala
```diff
@@ -76,13 +77,14 @@ class NotificationPageSingleElementFactory @Inject()(
...
-    notification.responseType match {
-      case ControlResponse        => controlResponseConverter.convert(notification)
-      case MovementTotalsResponse => movementTotalsResponseConverter.convert(notification)
-      case MovementResponse       => movementResponseConverter.convert(notification)
-      case _                      => buildForUnspecified(notification.timestampReceived)
-    }
+    converters
+      .find(_.canConvertFrom(notification))
+      .map(_.convert(notification))
+      .getOrElse(buildForUnspecified(notification.timestampReceived))
```
PatrykRudnicki: I think it'll be easier to stay with pattern matching.

```
(notification.responseType, notification.messageCode) match {
    case (ControlResponse, _) => ???
    case (MovementTotalsResponse, EMR.code) => ???
    case ... etc
}
```

test/models/viewmodels/notificationspage/NotificationPageSingleElementFactorySpec.scala
```diff
@@ -40,17 +42,19 @@ class NotificationPageSingleElementFactorySpec extends WordSpec with MustMatcher
   private val formatter = DateTimeFormatter.ISO_OFFSET_DATE_TIME.withZone(ZoneId.systemDefault())
   private val testTimestamp = ZonedDateTime.parse(testTimestampString, formatter).toInstant
 
-  private val crcCodeKeyFromDecoder = CrcCode.Success
-  private val roeKeyFromDecoder = RoeCode.DocumentaryControl
-  private val soeKeyFromDecoder = SoeCode.DeclarationAcceptance
+  private val crcKeyFromDecoder = CRCCode.Success
+  private val icsKeyFromDecoder = ICSCode.InvalidationByCustoms
+  private val roeKeyFromDecoder = ROECode.DocumentaryControl
+  private val soeKeyFromDecoder = SOECode.DeclarationAcceptance
```
PatrykRudnicki: `ducrSoeKeyFromDecoder`, below we have `mucr` to have consistent naming

MaciejRewera: Renamed.

app/models/viewmodels/notificationspage/NotificationPageSingleElementFactory.scala
bambuchaAdm: Here i am not sure - Maciek's version tails better story about what actual happen in reality - and IMO it is easier to read

MaciejRewera: Staying with pattern matching here would mean the logic leaking from converters to factory. Also, since the converters 'knows' what they can convert from, there is no point in doubling this part of the logic.

PR #99, opened 2019-09-23
CEDS-1694 EORI whitelist 

bambuchaAdm:
Including removing ip whitelisting


app/controllers/actions/AuthAction.scala
```diff
@@ -60,3 +70,12 @@ trait AuthAction
     extends ActionBuilder[AuthenticatedRequest, AnyContent] with ActionFunction[Request, AuthenticatedRequest]
 
 case class NoExternalId() extends NoActiveSession("No externalId was found")
+@ProvidedBy(classOf[EoriWhitelistProvider])
+class EoriWhitelist(values: Seq[String]) {
+  def allow(user: SignedInUser): Boolean = values.isEmpty || values.contains(user.eori)
+}
+
+class EoriWhitelistProvider @Inject()(configuration: Configuration) extends Provider[EoriWhitelist] {
```
SteveSugden: Wouldn't we normally inject our AppConfig class (rather than Configuration) and leave that class to lazy load the config values (i.e. eori list)?

conf/application.conf
```diff
@@ -132,10 +132,8 @@ auditing {
   }
 }
 
-ip-whitelist {
-  enabled: false
-  list: ["127.0.0.1"]
-  exclude: ["/ping/ping", "/healthcheck"]
+whitelist {
```
SteveSugden: Are we going to have some known eoris for local/dev/qa/staging config?

app/controllers/actions/AuthAction.scala
PatrykRudnicki: Approach with injecting `Configuration` give us possibility to split huge `AppConfig` to few different for specific Configuration. So for whitelisting we will have one small Provider

PatrykRudnicki: Moreover what you can see here, we don't need anything from `AppConfig` in `AuthAction`, just information about eori, so instead of injecting huge class we can inject this and have access to just one configuration

app/controllers/actions/AuthAction.scala
```diff
@@ -50,7 +56,11 @@ class AuthActionImpl @Inject()(override val authConnector: AuthConnector, mcc: M
 
           val cdsLoggedInUser = SignedInUser(eori.get.value, allEnrolments)
 
-          block(AuthenticatedRequest(request, cdsLoggedInUser))
+          if (eoriWhitelist.allow(cdsLoggedInUser)) {
```
ed0906: `allow` doesnt seem right to me, it implies it is "doing an action" of some kind.

I'd rename it to `contains` so it reads like "if eori whitelist contains CDS logged in user"

app/controllers/actions/AuthAction.scala
```diff
@@ -60,3 +70,12 @@ trait AuthAction
     extends ActionBuilder[AuthenticatedRequest, AnyContent] with ActionFunction[Request, AuthenticatedRequest]
 
 case class NoExternalId() extends NoActiveSession("No externalId was found")
+@ProvidedBy(classOf[EoriWhitelistProvider])
```
ed0906: Why do we need a Provider?? Just make it a Guice managed class

conf/application.conf
ed0906: I think he's done logic to allow everyone if the whitelist is empty

test/unit/controllers/actions/EoriWhitelistProviderSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+import play.api.Configuration
+import unit.base.UnitSpec
+
+class EoriWhitelistProviderSpec extends UnitSpec {
+
+  "EoriWhitelistProvider" should {
+    "create list from configuration" in {
+      val config = Configuration("whitelist.eori.0" -> "12345")
+      val provider = new EoriWhitelistProvider(config)
+      provider.get() mustBe a[EoriWhitelist]
```
ed0906: As above, I see no reason for a provider, just adds uneccessary code

app/controllers/actions/AuthAction.scala
ed0906: Personally, I don't see `AppConfig` being a problem, it's nice being able to look in once place for all configuration

conf/application.conf
SteveSugden: 👍 +1  Missed the "isEmpty" check

app/controllers/actions/AuthAction.scala
bambuchaAdm: `AppConfig` becomes [God Object](https://en.wikipedia.org/wiki/God_object) for system configuration. Every time you have use it it ends with mock or stub that configure half of system. Or become mocking `AppConfig` itself and to be honest i don't know, which is worse.

Generally if i would like to use `AppConfig` from level of provider there would be only to access configuration. Injecting object to access other object is quite [serious code smell (section digging into collabolators)](http://misko.hevery.com/code-reviewers-guide/)   On the other hand i can put new keys into `AppConfig` but this would break [SRP of `AppCofnig`](https://en.wikipedia.org/wiki/Single_responsibility_principle) and now we have multiple purpose why we have to change `AppConfig`. Using `AppConfig` would also violate [Demeter Law](https://en.wikipedia.org/wiki/Law_of_Demeter) expose internal structure and clutter this class into multiple place in code.

Here i don't have to use `ServicesConfig` and i don't see pros for using here `AppConfig` - it would just put more responsibilities on single object, which is now quite big and hard to maintain now, ultimately make it harder to test.

test/base/MockAuthConnector.scala
```diff
@@ -31,7 +31,8 @@ import scala.concurrent.Future
 
 trait MockAuthConnector extends MockitoSugar with Stubs {
   lazy val authConnectorMock: AuthConnector = mock[AuthConnector]
-  val mockAuthAction = new AuthActionImpl(authConnectorMock, stubMessagesControllerComponents())
+  val mockAuthAction =
+    new AuthActionImpl(authConnectorMock, new EoriWhitelist(Seq.empty), stubMessagesControllerComponents())
```
ed0906: This should be a mock in my opinion

app/controllers/actions/AuthAction.scala
ed0906: My proposal is:

```
class EoriWhitelist @Inject()(configuration: Configuration) {
  lazy val values = configuration.get[Seq[String]]("whitelist.eori")
  def contains(user: SignedInUser): Boolean = values.isEmpty || values.contains(user.eori)
}
```

PR #102, opened 2019-09-24
CEDS-1660 Audit Movements in Json format

PauloPortugal:
Keystore which is the tech underlying the CacheMap encrypts the data, so
we need to get each entry and then transform it into Json.
This is another reason for us to move away from Keystore

### DIFF
--- a/app/services/audit/AuditService.scala
+++ b/app/services/audit/AuditService.scala
@@ -17,11 +17,12 @@
 package services.audit
 
 import com.google.inject.Inject
-import forms.Choice
+import forms._
 import javax.inject.Named
 import play.api.Logger
 import play.api.libs.json.{JsObject, Json}
 import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.http.cache.client.CacheMap
 import uk.gov.hmrc.play.audit.AuditExtensions
 import uk.gov.hmrc.play.audit.http.connector.AuditResult.{Disabled, Failure, Success}
 import uk.gov.hmrc.play.audit.http.connector.{AuditConnector, AuditResult}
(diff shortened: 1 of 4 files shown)


app/services/audit/AuditService.scala
```diff
@@ -121,17 +122,36 @@ class AuditService @Inject()(connector: AuditConnector, @Named("appName") appNam
...
   }
 
+  def getMovementsData(choice: Choice, cacheMap: CacheMap): JsObject = {
+    val movementDetails =
+      if (choice.value == Choice.AllowedChoiceValues.Arrival)
+        Json.toJson(cacheMap.getEntry[ArrivalDetails](MovementDetails.formId))
+      else Json.toJson(cacheMap.getEntry[DepartureDetails](MovementDetails.formId))
+
+    val userInput = Map(
+      ConsignmentReferences.formId -> Json.toJson(
```
ed0906: This could be an opportunity to make an Arrival/Departure model....

app/services/audit/AuditService.scala
```diff
@@ -121,17 +122,36 @@ class AuditService @Inject()(connector: AuditConnector, @Named("appName") appNam
...
+        cacheMap.getEntry[ConsignmentReferences](ConsignmentReferences.formId)
+      ),
+      GoodsDeparted.formId -> Json.toJson(cacheMap.getEntry[GoodsDeparted](GoodsDeparted.formId)),
+      Location.formId -> Json.toJson(cacheMap.getEntry[Location](Location.formId)),
+      MovementDetails.formId -> movementDetails,
+      Transport.formId -> Json.toJson(cacheMap.getEntry[Transport](Transport.formId)),
+      ArrivalReference.formId -> Json.toJson(cacheMap.getEntry[ArrivalReference](ArrivalReference.formId))
+    )
+    Json.toJson(userInput).as[JsObject]
```
PatrykRudnicki: Is it `as[JsObject]` necessary? You'll have `JsObject` directly from `Json.toJson(...)`

PauloPortugal: This has been addressed.

PauloPortugal: Actually you need it, otherwise, you get a compilation error

PR #103, opened 2019-09-24
CEDS-1625 - add transportId to Transport page

SteveSugden:
- only show Transport page when departing out of UK
- update summary page (with new data and human-readable values)


app/controllers/MovementDetailsController.scala
```diff
@@ -88,8 +90,13 @@ class MovementDetailsController @Inject()(
       .fold(
         (formWithErrors: Form[DepartureDetails]) => Future.successful(Left(departureDetailsPage(formWithErrors))),
         validForm =>
-          customsCacheService.cache[DepartureDetails](movementCacheId, formId, validForm).map { _ =>
-            Right(controllers.routes.TransportController.displayPage())
+          customsCacheService.cache[DepartureDetails](movementCacheId, formId, validForm).map { cacheMap: CacheMap =>
+            cacheMap.getEntry[GoodsDeparted](GoodsDeparted.formId) match {
+              case (Some(goodsDeparted)) if (goodsDeparted.departedPlace == AllowedPlaces.outOfTheUk) =>
```
ed0906: No need for the brackets on this line around the `Some`

app/forms/GoodsDeparted.scala
```diff
@@ -29,8 +30,8 @@ object GoodsDeparted {
   val formId = "GoodsDeparted"
 
   object AllowedPlaces {
-    val outOfTheUk = "OutOfTheUk"
-    val backIntoTheUk = "BackIntoTheUk"
+    val outOfTheUk = "outOfTheUk"
```
ed0906: controversial

SteveSugden: Why?   The actual values aren't submitted anywhere and I wanted to re-use them to create message keys (at one point - not doing that any more so I could revert the above change if it upsets anyone).

app/forms/Transport.scala
```diff
@@ -42,15 +42,33 @@ object Transport {
...
+      .verifying("transport.transportId.empty", nonEmpty)
+      .verifying("transport.transportId.error", isEmpty or (noLongerThan(35) and isAlphanumeric))
   )(Transport.apply)(Transport.unapply)
 
-  def form(): Form[Transport] = Form(mapping)
+  def form: Form[Transport] =
+    Form(mapping)
+
+  def messageKey(mode: String) =
+    s"transport.modeOfTransport.${mode match {
```
bambuchaAdm: please extract as local function

app/forms/Transport.scala
```diff
@@ -42,15 +42,33 @@ object Transport {
...
+    "transportId" -> text()
+      .verifying("transport.transportId.empty", nonEmpty)
+      .verifying("transport.transportId.error", isEmpty or (noLongerThan(35) and isAlphanumeric))
   )(Transport.apply)(Transport.unapply)
 
-  def form(): Form[Transport] = Form(mapping)
+  def form: Form[Transport] =
+    Form(mapping)
+
+  def messageKey(mode: String) =
```
PatrykRudnicki: In my opinion this method is terrible

PatrykRudnicki: This `match` should be in separate method if we need something like this

PatrykRudnicki: It should be replaced by enum, case object or anything to make it better.

Without context I'm not able to understand what are we doing here and why

app/forms/Transport.scala
```diff
@@ -42,15 +42,33 @@ object Transport {
...
+    s"transport.modeOfTransport.${mode match {
+      case ModesOfTransport.Sea                => "sea"
+      case ModesOfTransport.Rail               => "rail"
+      case ModesOfTransport.Road               => "road"
+      case ModesOfTransport.Air                => "air"
+      case ModesOfTransport.PostalOrMail       => "postalOrMail"
+      case ModesOfTransport.FixedInstallations => "fixed"
+      case ModesOfTransport.InlandWaterway     => "inlandWaterway"
+      case ModesOfTransport.Other              => "other"
+      case _                                   => "unknown"
```
PatrykRudnicki: You're adding unknown, build whole message key `transport.modeOfTransport.unknown`

We don't have key like this and you didn't add it.

app/forms/GoodsDeparted.scala
```diff
@@ -43,4 +43,11 @@ object GoodsDeparted {
   )(GoodsDeparted.apply)(GoodsDeparted.unapply)
 
   def form(): Form[GoodsDeparted] = Form(mapping)
+
+  def messageKey(place: String) =
+    s"goodsDeparted.departedPlace.${place match {
```
bambuchaAdm: Please extract match result to local variable

test/forms/TransportSpec.scala
```diff
@@ -53,41 +55,54 @@ class TransportSpec extends BaseSpec {
...
+
+    "has correct message mapping" in {
+      Transport.messageKey(Sea) must be("transport.modeOfTransport.sea")
+      Transport.messageKey(Rail) must be("transport.modeOfTransport.rail")
+      Transport.messageKey(Road) must be("transport.modeOfTransport.road")
+      Transport.messageKey(Air) must be("transport.modeOfTransport.air")
+      Transport.messageKey(PostalOrMail) must be("transport.modeOfTransport.postalOrMail")
+      Transport.messageKey(FixedInstallations) must be("transport.modeOfTransport.fixed")
+      Transport.messageKey(InlandWaterway) must be("transport.modeOfTransport.inlandWaterway")
+      Transport.messageKey(Other) must be("transport.modeOfTransport.other")
```
PatrykRudnicki: Missing test case for unknown

test/views/DepartureSummaryViewSpec.scala
```diff
@@ -26,10 +29,15 @@ import views.html.summary.departure_summary_page
 
 class DepartureSummaryViewSpec extends UnitViewSpec with Stubs with Injector {
 
-  val cachedData = cacheMapData("EDL")
+  val cachedDataOutOfUk = cacheMapData("EDL")
+  val cachedDataBackIntoUk = cachedDataOutOfUk + (GoodsDeparted.formId -> Json.toJson(
```
PatrykRudnicki: IMO naming, `cachedDataBackIntoUk` that use inside `cachedDataOutOfUk` I'll use this `cacheMapData("EDL")` once again

app/forms/Transport.scala
SteveSugden: Terrible is a bit harsh.   I've had a couple of comments that the match should be extracted.  Seems excessive but I could do that.  And you are right, the problem here is to convert a 'value' to a sensible 'display text' which would have been much simpler if the values were "proper" enums.

app/forms/Transport.scala
SteveSugden: 'unknown' should never happen but it it did I think it would be preferable to the app blowing up with match exception.  I'll add a message key for 'unknown' as well

app/forms/Transport.scala
PatrykRudnicki: This method looks like method in the view, everything in the string and match inside the string. Something like this inside some specific view it can be easier to understand that you're building message keys, because something is different. In practice probably it'll be more readable to use messages key with numbers not strings, it's something for us, so `transport.modeOfTransport.1` without transformation keys. IMO if we will transform data, and make a lot of changes between values -> keys -> data or something like this it will make debugging really difficult if we'll have problem with this page

PR #106, opened 2019-09-26
CEDS-1625 - fix back link for arrivals summary page

SteveSugden:


test/views/DepartureSummaryViewSpec.scala
```diff
@@ -72,6 +72,16 @@ class DepartureSummaryViewSpec extends UnitViewSpec with Stubs with Injector {
       departureSummaryViewOut.getElementById("title").text() mustBe messages("summary.departure.title")
     }
 
+    "have correct back link for depart out of uk" in {
+
+      departureSummaryViewOut.getElementById("link-back").attr("href") must endWith("/transport")
```
PatrykRudnicki: Small comment, for url you can use `routes.controllers.TransportController.displayPage().url`

PatrykRudnicki: so in that case we can use mustBe not must endWith, but actual approach is correct too

SteveSugden: Yes.  That would be better.    I copied a similar test.

test/views/ArrivalReferenceViewSpec.scala
```diff
@@ -70,7 +71,7 @@ class ArrivalReferenceViewSpec extends UnitViewSpec with Stubs with Injector {
       val backButton = createView().getElementById("link-back")
 
       backButton.text() mustBe messages("site.back")
-      backButton.attr("href") must endWith("/consignment-references")
+      backButton.attr("href") mustBe routes.ConsignmentReferencesController.displayPage().url
```
bambuchaAdm: haveHref or something i coś takiego

PR #107, opened 2019-09-26
CEDS-1713 New implementation for choice

PatrykRudnicki:
Injecting execution context and body parsers in Actions to avoid injecting MessagesControllerComponents and use them from this trait.

Implemented new approach using sealed abstract class to give a possibility to use more readable messages in messages.en and don't create another class for form.
Moreover new approach is type safe and has validation during creating choice object.
Small boilerplate code shouldn't break readability.
    
Values for inventory linking exports request are build at the end and aren't used in different part of the codes.
    
Fixed tests to use prepared mocks not created just for one spec and removed duplication from it.
    
Those changes give us a possibility to use `arrival`, `departure` etc in cache naming, no more EDL, EAL.


app/forms/Choice.scala
```diff
@@ -18,32 +18,40 @@ package forms
...
 
-  import AllowedChoiceValues._
-  private val correctChoices = Set(Arrival, Departure, AssociateDUCR, DisassociateDUCR, ShutMucr, Submissions)
-
-  val choiceMapping: Mapping[Choice] = Forms.mapping(
-    "choice" -> requiredRadio("choicePage.input.error.empty")
-      .verifying("choicePage.input.error.incorrectValue", isContainedIn(correctChoices))
-  )(Choice.apply)(Choice.unapply)
+  val choiceMapping: Mapping[Choice] =
+    Forms.mapping("choice" -> requiredRadio("choicePage.input.error.empty"))(Choice.apply)(Choice.unapply)
```
bambuchaAdm: Here where we would try apply non existing choice we would get Exception from `form.bind` - please add validation step with some error. 

From perspective of UX - it better to have rerendered form with value with error then having "Internal Server Error" page

app/forms/Choice.scala
```diff
@@ -18,32 +18,40 @@ package forms
...
-    val AssociateDUCR = "Associate"
-    val DisassociateDUCR = "EAC"
-    val ShutMucr = "CST"
-    val Submissions = "SUB"
+  def unapply(status: Choice): Option[String] = Some(status.value)
+
+  def apply(input: String): Choice =
+    allChoices.find(_.value == input).getOrElse(throw new IllegalArgumentException("Incorrect choice"))
+
+  implicit object ChoiceValueFormat extends Format[Choice] {
```
SteveSugden: Much, much better than Strings.  But it still irks me that we need this much code to convert a string to a 'value'.  Roll-on proper enum support in scala!  :-)

app/metrics/MovementsMetrics.scala
```diff
@@ -18,27 +18,27 @@ package metrics
...
-    DisassociateDUCR -> metrics.defaultRegistry.counter(s"$disassociationMetric.counter"),
-    ShutMucr -> metrics.defaultRegistry.counter(s"$shutMucr.counter")
+    Arrival.value -> metrics.defaultRegistry.counter(s"$arrivalMetric.counter"),
+    Departure.value -> metrics.defaultRegistry.counter(s"$departureMetric.counter"),
+    AssociateDUCR.value -> metrics.defaultRegistry.counter(s"$associationMetric.counter"),
+    DisassociateDUCR.value -> metrics.defaultRegistry.counter(s"$disassociationMetric.counter"),
+    ShutMUCR.value -> metrics.defaultRegistry.counter(s"$shutMucr.counter")
   )
 
   def startTimer(feature: String): Context = timers(feature).time()
```
bambuchaAdm: We can have second overload method with `Choice` argument and then convert it internally to representation. This would increase confidence in API usage.

app/services/SubmissionService.scala
```diff
@@ -50,10 +50,10 @@ class SubmissionService @Inject()(
         auditService.auditAllPagesUserInput(choice, cacheMap)
 
         val movementAuditType =
-          if (choice.value == Choice.AllowedChoiceValues.Arrival) AuditTypes.AuditArrival else AuditTypes.AuditDeparture
+          if (choice == Arrival) AuditTypes.AuditArrival else AuditTypes.AuditDeparture
```
bambuchaAdm: Is AuditTypes is defined for every choice ?  If yes maybe would be nice to link it as Choice filed.

test/base/MockAuthConnector.scala
```diff
@@ -36,7 +39,7 @@ trait MockAuthConnector extends MockitoSugar with Stubs {
   val eoriWhitelistMock: EoriWhitelist = mock[EoriWhitelist]
 
   val mockAuthAction =
-    new AuthActionImpl(authConnectorMock, eoriWhitelistMock, stubMessagesControllerComponents())
+    new AuthActionImpl(authConnectorMock, eoriWhitelistMock, PlayBodyParsers()(NoMaterializer))(global)
```
bambuchaAdm: there is method for bodyParsers in `Helpers`

test/forms/ChoiceSpec.scala
```diff
@@ -26,20 +26,16 @@ class ChoiceSpec extends WordSpec with MustMatchers {
...
       }
 
-      "provided with a value not defined in AllowedChoiceValues" in {
-        val form = Choice.form().bind(incorrectChoiceJSON)
-
-        form.hasErrors must be(true)
-        form.errors.length must equal(1)
-        form.errors.head.message must equal("choicePage.input.error.incorrectValue")
+      "provided with an incorrect value" in {
+        intercept[IllegalArgumentException](Choice.form().bind(incorrectChoiceJSON))
```
bambuchaAdm: I think original version with form with errors is better rather then throwing exceptions

test/services/SubmissionServiceSpec.scala
```diff
@@ -41,42 +41,40 @@ import scala.concurrent.Future
...
 
   val mockAuditService = mock[AuditService]
-  val customsCacheServiceMock = mock[CustomsCacheService]
-  val customsExportsMovementConnectorMock = mock[CustomsDeclareExportsMovementsConnector]
 
   val submissionService =
     new SubmissionService(
-      customsCacheServiceMock,
-      customsExportsMovementConnectorMock,
+      mockCustomsCacheService,
```
bambuchaAdm: remove mock prefix - as result of techtro

test/testdata/MovementsTestData.scala
```diff
@@ -47,9 +47,9 @@ object MovementsTestData {
   val date = Date(Some(10), Some(8), Some(2018))
   val departureDetails = DepartureDetails(date)
 
-  def arrivalDepartureTimes(movementType: String): JsValue = movementType match {
-    case "EAL" => Json.toJson(ArrivalDetails(date, Time(Some("13"), Some("34"))))
-    case _     => Json.toJson(DepartureDetails(date))
+  def arrivalDepartureTimes(movementType: Choice): JsValue = movementType match {
+    case Arrival => Json.toJson(ArrivalDetails(date, Time(Some("13"), Some("34"))))
+    case _       => Json.toJson(DepartureDetails(date))
```
bambuchaAdm: This methods looks

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -87,23 +87,21 @@ class ChoiceControllerSpec extends ControllerSpec with OptionValues with BeforeA
...
 
       "form is incorrect" in {
 
-        val incorrectForm = Json.toJson(Choice("incorrect"))
-
-        val result = controller.submitChoice()(postRequest(incorrectForm))
-
-        status(result) mustBe BAD_REQUEST
+        intercept[IllegalArgumentException] {
+          await(controller.submitChoice()(postRequest(Json.toJson(Choice("incorrect")))))
```
bambuchaAdm: same - not exception but badRequest od as @PauloPortugal point [422 Unprocessable Entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422)

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -87,23 +87,21 @@ class ChoiceControllerSpec extends ControllerSpec with OptionValues with BeforeA
...
+        }
       }
     }
 
     "return 303 (SEE_OTHER)" when {
 
       "choice is Arrival" in {
 
-        val arrivalForm = JsObject(Map("choice" -> JsString(Arrival)))
+        val arrivalForm = JsObject(Map("choice" -> JsString(Arrival.value)))
```
bambuchaAdm: I am not 100% sure but if you had provided `Format` for `Choice` syntax like
```
Json.obj("choice" -> Arrival)
```
should works

test/views/ChoiceViewSpec.scala
```diff
@@ -107,20 +107,13 @@ class ChoiceViewSpec extends ViewSpec with ChoiceMessages with CommonMessages {
 
       getElementByCss(view, "#error-message-choice-input").text() must be(messages(choiceEmpty))
     }
-
-    "display error when choice is incorrect" in {
```
bambuchaAdm: as earlier please back to original logic

test/views/MovementConfirmationViewSpec.scala
```diff
@@ -25,19 +26,19 @@ import views.html.movement_confirmation_page
...
 
     "have correct messages" in {
 
       val messages = instanceOf[MessagesApi].preferred(request)
 
-      messages("movement.EAL.confirmation") mustBe "Arrival confirmation"
-      messages("movement.EDL.confirmation") mustBe "Departure confirmation"
-      messages("movement.choice.EAL.label") mustBe "Arrival"
-      messages("movement.choice.EDL.label") mustBe "Departure"
+      messages("movement.arrival.confirmation") mustBe "Arrival confirmation"
```
bambuchaAdm: Please move this to test key presence not value of them

app/forms/Choice.scala
bambuchaAdm: There is very good support but in third party library - https://github.com/lloydmeta/enumeratum - But here we hit the problem of type system expressiveness and type class derivation.

app/forms/Choice.scala
PatrykRudnicki: Fixed

app/metrics/MovementsMetrics.scala
PatrykRudnicki: Fixed

app/services/SubmissionService.scala
PatrykRudnicki: Nope, it's defined, but the logic for it is just related with Arrival and Departure

test/base/MockAuthConnector.scala
PatrykRudnicki: That doesn't work, stub is not enough because somehow we're using this. To use methods from Helpers we will need to add logic to stub that we take from this class. This approach is easier and less problematic.

if there is some reason where this approach is not good I can revert those changes.

test/forms/ChoiceSpec.scala
PatrykRudnicki: Fixed

test/services/SubmissionServiceSpec.scala
PatrykRudnicki: It's old mock, renaming this will affect all places where we're using this mock. I prefer to leave it alone for now.

test/testdata/MovementsTestData.scala
PatrykRudnicki: MovementTestData, moreover it's old implementation

test/unit/controllers/ChoiceControllerSpec.scala
PatrykRudnicki: Fixed

test/unit/controllers/ChoiceControllerSpec.scala
PatrykRudnicki: It's not working, you need to pass String

test/views/ChoiceViewSpec.scala
PatrykRudnicki: Fixed

test/views/MovementConfirmationViewSpec.scala
PatrykRudnicki: Next PR, we need to change it for all view specs. We don't have helpers method for it in this repository so I don't want to start this now

PatrykRudnicki:
@bambuchaAdm Please review it once again, All comments fixed

test/base/MockFactory.scala
```diff
@@ -104,7 +104,7 @@ object MockFactory extends MockitoSugar {
   def buildMovementsMetricsMock: MovementsMetrics = {
     val movementsMetricsMock = mock[MovementsMetrics]
 
-    when(movementsMetricsMock.startTimer(any())).thenReturn(new Timer().time())
+    when(movementsMetricsMock.startTimer(anyString)).thenReturn(new Timer().time())
```
bambuchaAdm: You also can put type paramter to any `any[String]()`

PatrykRudnicki: Just different approach

PR #108, opened 2019-09-26
CEDS-1662 ile errors from file

MaciejRewera:
This PR aligns the way we handle ILE and CHIEF errors.
As was agreed by the team, it also introduces CSV config files with messages keys. So the description for the errors to be displayed to the user is in messages.* files.


app/models/viewmodels/notificationspage/converters/ControlResponseConverter.scala
```diff
@@ -54,27 +54,17 @@ class ControlResponseConverter @Inject()(decoder: Decoder, dateTimeFormatter: Da
...
       case None =>
         logger.info(s"Received inventoryLinkingControlResponse with unknown error code: $errorCode")
         None
       case error => error
     }
 
-  private def buildCHIEFErrorExplanation(errorCode: String): Option[String] =
-    decoder.chiefErrorCode(errorCode).map(error => paragraph(error.description))
-
   private val paragraph: String => String = (text: String) => s"<p>$text</p>"
```
SteveSugden: Not sure about building HTML outside the views, but I guess you've not changed that.  I'd just have this converter return messages (or message keys).  But it's not important for this PR

app/models/viewmodels/notificationspage/converters/ControlResponseConverter.scala
```diff
@@ -54,27 +54,17 @@ class ControlResponseConverter @Inject()(decoder: Decoder, dateTimeFormatter: Da
...
+    decoder.actionCode(actionCode).map(code => paragraph(messages(code.messageKey)))
 
   // TODO move logging for missing error codes to backend
-  private def buildILEErrorExplanation(errorCode: String)(implicit messages: Messages): Option[String] =
-    decoder.ileErrorCode(errorCode).map(error => paragraph(messages(error.contentKey))) match {
+  private def buildErrorExplanation(errorCode: String)(implicit messages: Messages): Option[String] =
+    decoder.error(errorCode).map(error => paragraph(messages(error.messageKey))) match {
       case None =>
         logger.info(s"Received inventoryLinkingControlResponse with unknown error code: $errorCode")
         None
```
SteveSugden: Maybe return just the "raw" error code if it can't be "translated"?   Or a canned message like "Unknown error XXX" .   Then the user knows there was an error and can report it  (and someone can investigate and possibly update the csv file).

In which case, doesn't need to return Option.

MaciejRewera: This was actually the requirement from the business. Not to display anything for unknown errors. We are only logging this, which by the way is going to be moved to the BE. And then the logging level possibly changed to `warn`.

app/models/viewmodels/notificationspage/converters/ControlResponseConverter.scala
MaciejRewera: I agree, but I feel it is simpler to do it this way for now since the design is changing a lot in this area. And sometimes we need to use paragraphs, and sometimes we need lists...

app/models/viewmodels/notificationspage/converters/MovementResponseConverter.scala
```diff
@@ -51,7 +51,7 @@ class MovementResponseConverter @Inject()(decoder: Decoder, dateTimeFormatter: D
 
   private def buildCrcCodeExplanation(crcCode: String)(implicit messages: Messages): Option[String] = {
     val CrcCodeHeader = messages("notifications.elem.content.inventoryLinkingMovementResponse.crc")
-    decoder.crc(crcCode).map(code => paragraph(s"$CrcCodeHeader ${code.contentKey}"))
+    decoder.crc(crcCode).map(code => paragraph(s"$CrcCodeHeader ${code.messageKey}"))
```
bambuchaAdm: missing messages ?

conf/code_lists/chief_errors.csv
```diff
@@ -0,0 +1,25 @@
+E3481,decoder.chiefError.MucrAlreadyShut
```
bambuchaAdm: why decoder prefix ?

app/models/viewmodels/decoder/CHIEFError.scala
```diff
@@ -21,28 +21,30 @@ import play.api.Logger
...
+  * @param messageKey description for this error code
+  */
+case class CHIEFError(override val code: String, override val messageKey: String) extends CodeWithMessageKey
 
 object CHIEFError {
 
   private val logger = Logger(this.getClass)
-
-  def apply(code: String): Option[CHIEFError] = chiefErrors.find(_.code == code)
+  private val sourcePath = "code_lists/chief_errors.csv"
```
PatrykRudnicki: Are you sure about this sourcePath?

conf/code_lists/chief_errors.csv
bambuchaAdm: What do you think about `error.chief.MucrAlreadyShut` - tells more what it is and where it comes.

test/models/viewmodels/decoder/ILEErrorSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+import org.scalatest.{MustMatchers, WordSpec}
+
+class ILEErrorSpec extends WordSpec with MustMatchers {
+
+  "ILE Error" should {
+
+    "have correct amount of codes" in {
+
+      val expectedCodesAmount = 29
+      ILEError.allErrors.size mustBe expectedCodesAmount
```
bambuchaAdm: I don't know about that test - it would failed if something change - it good and bad depending from point of view. Let stay with this test for now.

test/models/viewmodels/decoder/ILEErrorSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    "contain non-empty code and description for every error" in {
+
+      ILEError.allErrors.map { error =>
+        error.code mustNot be(empty)
```
bambuchaAdm: :+1:

test/models/viewmodels/decoder/ILEErrorSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+        error.messageKey mustNot be(empty)
+      }
+    }
+
+    "contain correct prefix for all message keys" in {
+
+      val expectedPrefix = "decoder.ileError."
+
+      ILEError.allErrors.map { error =>
```
bambuchaAdm: maybe use of `foreach` would be more natural to read

app/models/viewmodels/notificationspage/converters/MovementResponseConverter.scala
MaciejRewera: This is being fixed in CEDS-1736. I am preparing PR for this today.

app/models/viewmodels/decoder/CHIEFError.scala
MaciejRewera: I've changed that in one of the commits. I moved all files for countries and errors to a separate folder.

conf/code_lists/chief_errors.csv
MaciejRewera: 👍

PR #111, opened 2019-09-30
CEDS-1662 display multiple errors

MaciejRewera:


app/models/viewmodels/notificationspage/converters/ResponseConverterProvider.scala
```diff
@@ -0,0 +1,68 @@
...
+  ): Try[NotificationPageSingleElementConverter] =
+    Try(notification.responseType match {
+      case MovementTotalsResponse => getMovementTotalsResponseConverter(notification)
+      case ControlResponse        => getControlResponseConverter(notification)
+      case MovementResponse       => movementResponseConverter
+    })
+
+  private def getMovementTotalsResponseConverter(
+    notification: NotificationFrontendModel
+  ): NotificationPageSingleElementConverter = notification.messageCode match {
```
PauloPortugal: :+1:

PauloPortugal: I imagine that we would not get another message code apart from these two.

app/models/viewmodels/notificationspage/NotificationPageSingleElementFactory.scala
```diff
@@ -77,22 +70,9 @@ class NotificationPageSingleElementFactory @Inject()(
...
-  private def buildForUnspecified(
-    responseTimestamp: Instant
-  )(implicit messages: Messages): NotificationsPageSingleElement =
-    NotificationsPageSingleElement(
-      title = messages("notifications.elem.title.unspecified"),
-      timestampInfo = dateTimeFormatter.format(responseTimestamp),
-      content = HtmlFormat.empty
-    )
+  def build(notification: NotificationFrontendModel)(implicit messages: Messages): NotificationsPageSingleElement = {
+    val responseConverter = responseConverterProvider.provideResponseConverter(notification)
```
PauloPortugal: :+1:

app/models/viewmodels/notificationspage/converters/ResponseConverterProvider.scala
MaciejRewera: For MovementTotalsResponse we expect only those two. In case we receive a different one, we eventually return `UnknownResponseConverter` instance.

PR #114, opened 2019-10-03
CEDS-1693 start page content

MaciejRewera:


app/assets/stylesheets/partials/_lists.scss
```diff
@@ -6,3 +6,18 @@
 .notification-error-explanation:only-child {
   list-style: none;
 }
+
+ul.dashed {
```
bambuchaAdm: Why tag name?

app/assets/stylesheets/partials/_lists.scss
```diff
@@ -6,3 +6,18 @@
...
+  list-style: none;
+  padding-left: 0;
+}
+
+ul.dashed > li {
```
bambuchaAdm: Are you sure we would like have direct Child here? Also usage of tag in selector? 

Maybe adding class nam on that `li` tag

app/views/start_page.scala.html
```diff
@@ -20,35 +20,75 @@
...
-                <h2 class="heading-medium">@messages("startPage.referenceTitle")</h2>
-                <p>
-                    @messages("startPage.reference") <b>@messages("startPage.reference.number")</b> <br>
-                @messages("startPage.reference.text")
-                </p>
-                <p><a href="https://www.gov.uk/government/organisations/hm-revenue-customs/contact/customs-international-trade-and-excise-enquiries" target="_blank" rel="noopener noreferrer">@messages("startPage.enquiries")</a></p>
-            </div>
-        </div>
+
+  <div class="govuk-grid-row">
```
ed0906: I dont think `govuk-...` classes do anything, unless you've added them.

The grid system we use is on https://govuk-elements.herokuapp.com/layout/

app/views/start_page.scala.html
```diff
@@ -20,35 +20,75 @@
 
 @()(implicit request: Request[_], messages: Messages)
 
+@customsDeclarationsOutOfEuUrl = @{"https://www.gov.uk/guidance/customs-declarations-for-goods-taken-out-of-the-eu"}
```
bambuchaAdm: Looks like Configuration urls

app/views/start_page.scala.html
MaciejRewera: This might be the reason that the content is slightly narrower than previously... Fixing it.

app/views/start_page.scala.html
```diff
@@ -20,35 +20,75 @@
...
-        </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      @components.heading(
+        sectionHeader = messages("startPage.title.sectionHeader"),
+        title = messages("startPage.title")
+      )
+    </div>
+    <div class="govuk-grid-column-two-thirds">
```
bambuchaAdm: We have two times 2/3 which is not SUM up to one? Intentional?

app/views/start_page.scala.html
```diff
@@ -20,35 +20,75 @@
...
+      </p>
     </div>
+
+    <br/>
+    <a id="button-start" class="button button-start" href="@routes.ChoiceController.displayChoiceForm" role="button">
+      @messages("startPage.buttonName")
+    </a>
+
+    <div>
+      <br/>
```
bambuchaAdm: Do we have to get those 4 brs? Can we have 4 em top margin?

test/views/StartViewSpec.scala
```diff
@@ -16,190 +16,127 @@
...
-
-        getElementByCss(createView(), "article>div>div>ul>li:nth-child(7)")
-          .text() must be(messages(listItemSevenUrl) + " " + messages(listItemSeven))
-      }
-
-      "link in seventh element to\"Customs procedure codes\" page" in {
-
-        val link =
-          getElementByCss(createView(), "article>div>div>ul>li:nth-child(7)>a")
+      view.getElementById("contents-element-1").text() mustBe "startPage.beforeYouStart.header"
```
bambuchaAdm: Are we interest with structure or just fact that was rendered in cirrect section.

MaciejRewera: As I remember we were testing the structure some time ago, but moved away from that.

app/views/start_page.scala.html
MaciejRewera: This class is already changed to `column-full`

app/views/start_page.scala.html
MaciejRewera: Good point. Changed.

app/views/start_page.scala.html
MaciejRewera: Added those to config

app/views/start_page.scala.html
```diff
@@ -14,14 +14,12 @@
...
+@import config.AppConfig
 @import controllers.routes
 
 @this(main_template: views.html.main_template)
 
-@()(implicit request: Request[_], messages: Messages)
-
-@customsDeclarationsOutOfEuUrl = {"https://www.gov.uk/guidance/customs-declarations-for-goods-taken-out-of-the-eu"}
-@serviceAvailabilityUrl = {"https://www.gov.uk/guidance/customs-declaration-service-service-availability-and-issues"}
+@()(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
```
bambuchaAdm: why not inject app config directly to view ?

app/views/start_page.scala.html
```diff
@@ -76,11 +74,7 @@ <h2 id="report-your-arrival-and-departure" class="heading-medium">@messages("sta
         @messages("startPage.buttonName")
       </a>
 
-      <div>
-        <br/>
-        <br/>
-        <br/>
-        <br/>
+      <div style="margin-top: 5em">
```
bambuchaAdm: why inline style ?

app/views/start_page.scala.html
MaciejRewera: Did not think about it. It is injected now.

app/views/start_page.scala.html
MaciejRewera: There was no scss file with layouts but it might be a good time to create one.

PR #115, opened 2019-10-04
CEDS-1766 Move Arrive/Depart submission to the backend

PauloPortugal:


app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -65,22 +64,22 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
   )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] =
     postRequest(DucrDisassociation, requestXml)
 
-  def sendShutMucrRequest(requestXml: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] =
-    postRequest(ShutMucr, requestXml)
-
   private def postRequest(
     actionType: ActionType,
-    requestXml: String
+    payload: String
```
bambuchaAdm: from my perspective we should have her as much detailed type as we can - here we would be able to POST "fizzbuzz" and type system would not prevent before it

app/models/requests/MovementDetailsRequest.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.requests
+import play.api.libs.json.{Json, OFormat}
+
+case class MovementDetailsRequest(dateTime: String)
```
bambuchaAdm: Why DateTime is string ?

app/models/requests/MovementRequest.scala
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package models.requests
+
+import forms._
+import play.api.libs.json.{Json, OFormat}
+
+case class MovementRequest(
+  choice: String,
```
bambuchaAdm: Why do you use String here ?

app/services/SubmissionService.scala
```diff
@@ -66,11 +67,11 @@ class SubmissionService @Inject()(
 
   private def sendMovementRequest(
     choice: Choice,
-    data: InventoryLinkingMovementRequest
+    data: MovementRequest
   )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] =
     choice match {
-      case Arrival   => connector.sendArrivalDeclaration(data.toXml)
-      case Departure => connector.sendDepartureDeclaration(data.toXml)
+      case Arrival   => connector.sendArrivalDeclaration(Json.toJson(data).toString())
```
bambuchaAdm: Json.toString is asking for trouble

app/connectors/CustomsDeclareExportsMovementsConnector.scala
ed0906: Why are we using `PostString`? its not xml any more so we can post JSON

app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -65,22 +64,22 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
...
+      .POSTString[HttpResponse](movementSubmissionUrl(actionType), payload, CommonMovementsHeaders)
       .andThen {
         case Success(response) =>
           logger.debug(s"CUSTOMS_DECLARE_EXPORTS_MOVEMENTS response on ${actionType.value}. $response")
         case Failure(exception) =>
           logger.warn(s"CUSTOMS_DECLARE_EXPORTS_MOVEMENTS failure on ${actionType.value}. $exception ")
       }
 
+  def sendShutMucrRequest(requestXml: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] =
+    postRequest(ShutMucr, requestXml)
```
ed0906: The point of this was to get rid of the xml from the front end :)
Can we change this to submit the MovementRequest

app/models/requests/MovementDetailsRequest.scala
ed0906: `Instant` would be better :)

app/models/requests/MovementRequest.scala
ed0906: Would be nicer as a `MovementType` enum

ed0906: Would force that it can only be ARRIVE or DEPART

app/services/SubmissionService.scala
ed0906: Agreed, can we pass the model to the connector please

app/models/requests/MovementDetailsRequest.scala
PauloPortugal: We do not need to send an instant, because we are not storing this information to the DB. All we are doing is converting the data we collected from the FE pages, which will return a String and then submit that to the BE and the BE will relay. that messages (converting into XML).
Yes, I would otherwise convert it to Instant if there was a need.

app/connectors/CustomsDeclareExportsMovementsConnector.scala
PauloPortugal: This is the work @PatrykRudnicki is working on and I did not want to change it, to avoid code conflicts. This has been already removed by his work

app/connectors/CustomsDeclareExportsMovementsConnector.scala
PauloPortugal: :man_facepalming: Good point.

app/services/SubmissionService.scala
PauloPortugal: Done

app/models/requests/MovementRequest.scala
PauloPortugal: Done

PR #116, opened 2019-10-04
CEDS-1767 Switch to new endpoint for consolidations

PatrykRudnicki:


app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -31,47 +32,42 @@ import scala.concurrent.{ExecutionContext, Future}
...
-  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] = postRequest(DucrAssociation, requestXml)
+  def sendDepartureDeclaration(requestXml: String)(implicit hc: HeaderCarrier): Future[HttpResponse] =
+    postRequest(Departure, requestXml)
 
-  def sendDisassociationRequest(
-    requestXml: String
-  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] =
-    postRequest(DucrDisassociation, requestXml)
+  def sendConsolidationRequest(consolidation: ConsolidationRequest)(implicit hc: HeaderCarrier): Future[HttpResponse] =
+    httpClient.POST[ConsolidationRequest, HttpResponse](
```
ed0906: return type `HttpResponse`?

ed0906: Means its leaking Connector code to the Service

ed0906: I'd just return ConsolidationRequest, but call it `Consolidation` or something

PatrykRudnicki: Just for now I returned HttpResponse, because our backend doesn't return ConsolidationRequest actually, but Consolidation, another class that we need to build in the frontend to be consistent. As I discussed with you we will return Request from the backend but I don't want to break actual implementation. I can prepare parallel PR for movements to return ConsoldationRequest and change it here. It'll be better or you have different proposal?

app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -31,47 +32,42 @@ import scala.concurrent.{ExecutionContext, Future}
...
-  def sendDepartureDeclaration(
-    requestXml: String
-  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] = postRequest(Departure, requestXml)
+  def sendArrivalDeclaration(requestXml: String)(implicit hc: HeaderCarrier): Future[HttpResponse] =
+    postRequest(Arrival, requestXml)
 
-  def sendAssociationRequest(
-    requestXml: String
-  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] = postRequest(DucrAssociation, requestXml)
+  def sendDepartureDeclaration(requestXml: String)(implicit hc: HeaderCarrier): Future[HttpResponse] =
```
bambuchaAdm: parameter name ?

PatrykRudnicki: departure and arrival is still based on the xml

PR #117, opened 2019-10-04
CEDS-1693 start page polishing

MaciejRewera:
Just some minor changes to align with Declarations code

### DIFF
--- a/app/assets/stylesheets/partials/_layout.scss
+++ b/app/assets/stylesheets/partials/_layout.scss
@@ -1,4 +1,4 @@
 
-.back-to-top-panel {
-  margin-top: 5em
+.mt-5 {
+  margin-top: 5rem;
 }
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/partials/_layout.scss
```diff
@@ -1,4 +1,4 @@
 
-.back-to-top-panel {
-  margin-top: 5em
-}
\ No newline at end of file
+.mt-5 {
+  margin-top: 5rem;
```
PatrykRudnicki: `rem`? Should it be `em`?

MaciejRewera: `em` sets the size relatively to the container it is used in. `rem` is relative to a base size. So  it is more predictable to use the latter.

PR #119, opened 2019-10-07
CEDS-1767 Handling ConsolidationRequest as response in connector

PatrykRudnicki:


app/models/external/requests/ConsolidationType.scala
```diff
@@ -0,0 +1,29 @@
...
+import play.api.libs.json.{Format, Reads, Writes}
+
+object ConsolidationType extends Enumeration {
+  type ConsolidationType = Value
+
+  val ASSOCIATE_DUCR = Value("EAC")
+  val DISASSOCIATE_DUCR = Value("EAC")
+  val SHUT_MUCR = Value("CST")
+
+  implicit val format = Format(Reads.enumNameReads(ConsolidationType), Writes.enumNameWrites)
```
bambuchaAdm: I have feeling that this should be not enumName but enumValue or something simillar

PatrykRudnicki: It building correct Json, it's consistent with backend. Moreover, there is not value now

PR #121, opened 2019-10-09
CEDS-1725

MaciejRewera:
In order to be able to quickly switch between different 'types' of explanations there is a new feature flag added to config. Based on that correct suffix for messages keys is choosen.
For this purpose we need to store both versions.


app/models/viewmodels/notificationspage/ResponseErrorExplanationSuffixProvider.scala
```diff
@@ -0,0 +1,31 @@
...
+import config.AppConfig
+import javax.inject.Inject
+
+class ResponseErrorExplanationSuffixProvider @Inject()(appConfig: AppConfig) {
+
+  private val DefaultSuffix = ".Exports"
+
+  lazy val suffix: String = appConfig.responseErrorExplanationMode match {
+    case "CDS"     => ".CDS"
+    case "Exports" => ".Exports"
```
PatrykRudnicki: This case is unnecessary. Default value cover it.

conf/messages.en
```diff
@@ -410,60 +410,116 @@ decoder.soe.ConsolidationOpen = the MUCR is still open, or contains no declarati
...
-error.chief.ArrivalAlreadyNotified = Anticipated arrival not allowed - Arrival already notified
-error.chief.InvalidMrnFormat = MRN: Format is invalid
-error.chief.MrnNotMatchEntry = MRN: Does not match the latest version on the Entry
-error.chief.CannotArriveContainedMucr = Unable to arrive MUCR as it is contained in another MUCR
-error.chief.CannotDepartContainedMucr = Unable to depart MUCR as it is contained in another MUCR
-error.chief.NotSuitableEntryReference = Entry Reference is not suitable for Export movement
-error.chief.MucrNotShutForAssociation = Only shut MUCRs can be associated into another MUCR
-error.chief.CannotShutMovingConsolidation = Cannot shut the consolidation since it is moving
+error.ile.InvalidUcrFormat.CDS = Invalid UCR Format, DUCR or MUCR
+error.ile.ClientIdValidationFailed.CDS = Client ID validation fails
```
PatrykRudnicki: Still I'll think that in the file it will be easier. CSV with three columns `errorCode`, `exportsMessage`, `CDSMessage` message, load everything at the beginning of application and just based on the flag take CDS or Exports message, but it just different approach.

Approach with messages created manually can causes typos in message keys and it's more difficult to maintain.

app/models/viewmodels/notificationspage/ResponseErrorExplanationSuffixProvider.scala
SteveSugden: Why not put the 'dot' in the config value and just use that (i.e. no need for this provider class).  Unless you want to prevent spreading the AppConfig around.

Alternatively, this provider can just return s".${appConfig.responseErrorExplanationMode}"

conf/application.conf
```diff
@@ -93,10 +93,12 @@ microservice {
     }
 
     features {
+      list-of-available-response-error-explanation-modes: "CDS, Exports"
```
SteveSugden: Is this config value used?

conf/messages.en
SteveSugden: ... but it's easier to translate (and we don't have to worry about it).   Do we have a test somewhere that every error code in the error code file has an entry in the messages file?

PatrykRudnicki: yeah, that's true

app/models/viewmodels/notificationspage/ResponseErrorExplanationSuffixProvider.scala
MaciejRewera: I know it is not necessarty, but this is more about documenting what do we expect. In this case it is just a coincidence that the default value is the same.

conf/application.conf
MaciejRewera: It's not, but I found a similar solution in https://github.com/hmrc/app-config-qa/commit/b75be60f7165d4c2b1de101b217b2936d9769049#diff-88ec3c9b73885d4eacf1d4667e69651e
This is more of an info for someone changing the config, what values they can use.

PatrykRudnicki: I think it can be removed, because journey we need to have different based on the environment, it's affect user point of view because he will see more choices or less. Errors are our internal stuff. We're not using this and the point of showing specific error is based on the feature flag.

I don't want to keep this, because if it's unused it can be difficult to find the reason of creation in few weeks or it will be difficult to decide to remove it because it can be somehow used. I remember last ticket from @SteveSugden where it was difficult to decide what part we can remove

app/models/viewmodels/notificationspage/ResponseErrorExplanationSuffixProvider.scala
```diff
@@ -0,0 +1,31 @@
...
+    case _         => DefaultSuffix
```
bambuchaAdm: Maybe some info logger about that fact would be nice addition. This would be definitely helpful on debugging deployment in future.

test/models/viewmodels/decoder/CHIEFErrorSpec.scala
```diff
@@ -38,12 +41,39 @@ class CHIEFErrorSpec extends UnitSpec {
...
-
-      val expectedPrefix = "error.chief."
-
-      CHIEFError.allErrors.foreach { error =>
-        error.messageKey must include(expectedPrefix)
+    "have translations for all errors" in {
+      val messages = UnitViewSpec.realMessagesApi.preferred(FakeRequest().withCSRFToken)
+
+      val chiefErrorsNames = Seq(
+        "MucrAlreadyShut",
```
SteveSugden: Nit pick - couldn't we read these values out of the csv file(s)?   That way we ensure the message file is in sync without having to change the test.

PR #125, opened 2019-10-11
CEDS-1764 MUCR/DUCR validation

SteveSugden:
Separates out the MUCR and DUCR parts of the previous validation.  I.e. we now treat these references as different when validating


app/forms/ConsignmentReferences.scala
```diff
@@ -38,14 +39,41 @@ object ConsignmentReferences {
...
+        .verifying("consignmentReferences.reference.value.empty", nonEmpty)
+    )(ConsignmentReferences.apply)(ConsignmentReferences.unapply)
 
   def form(): Form[ConsignmentReferences] = Form(mapping)
 
 }
+
+object ConsignmentReferencesForm {
+
+  def bindFromRequest(implicit request: play.api.mvc.Request[_]) = {
```
ed0906: 👍

ed0906: Any reason why there isnt an import for `play.api.mvc.Request` rather than having the package prefix?

app/forms/ConsignmentReferences.scala
```diff
@@ -38,14 +39,41 @@ object ConsignmentReferences {
...
+    val baseForm = ConsignmentReferences.form().bindFromRequest
+
+    if (baseForm.errors.nonEmpty) {
+      baseForm
+    } else {
+      baseForm.value
+        .map(
+          reference =>
+            if (referenceValid(reference)) { baseForm } else {
```
PatrykRudnicki: Those `{ }` are unnecessary

app/forms/ConsignmentReferences.scala
PatrykRudnicki: My propose is to specify `AnyContent` then using `_`

app/forms/ConsignmentReferences.scala
```diff
@@ -38,14 +39,41 @@ object ConsignmentReferences {
...
+              baseForm.withError("referenceValue", "consignmentReferences.reference.value.error")
+          }
+        )
+        .getOrElse(baseForm)
```
PatrykRudnicki: A lot of logic here, I think we'll be able to make it shorter or some kind easier. Why this reference can't be validated in the mapping?

You can use verifying when you build whole object, not just for elements

SteveSugden: I can be validated in the mapping.  But you can't then include the key to the field that's wrong (and therefore it doesn't get highlighted)

app/forms/ConsignmentReferences.scala
SteveSugden: Fixed.  Happy now @ed0906 ?

PR #126, opened 2019-10-11
CEDS-1085 Add error table to Choice page

MaciejRewera:


app/views/components/error_summary.scala.html
```diff
@@ -26,7 +26,7 @@ <h2 class="h2-heading" id="error-summary-heading">
 
         <ul role="list" class="error-summary-list">
             @for(error <- errors) {
-                <li><a href="#@{error.key.replace(".", "_").replace('[', '_').replace("]", "")}">@messages(error.message, error.args:_*)</a></li>
+                <li><a id="@{error.key}-error" href="#@{error.key.replace(".", "_").replace('[', '_').replace("]", "")}">@messages(error.message, error.args:_*)</a></li>
```
ed0906: We should replace the `.`s before using the key in an ID

ed0906: You can create a variable with
```
@defining(error..key.replace(".", "_")) { id => 
  // some html
}
```

If you need to use it in the href too

test/views/ChoiceViewSpec.scala
```diff
@@ -74,28 +85,28 @@ class ChoiceViewSpec extends ViewSpec with ChoiceMessages with CommonMessages {
...
+      ensureRadioIsUnChecked(view, "disassociate")
+      ensureRadioIsUnChecked(view, "shut_mucr")
     }
 
     "display 'Save and continue' button on page" in {
 
       val view = createView()
 
-      val saveButton = getElementByCss(view, "#submit")
+      val saveButton = view.select("#submit")
```
ed0906: `view.getElementById("submit")`

test/views/ChoiceViewSpec.scala
```diff
@@ -106,14 +117,20 @@ class ChoiceViewSpec extends ViewSpec with ChoiceMessages with CommonMessages {
...
 
     "display error when choice is incorrect" in {
 
       val view = createView(Choice.form().bind(Map("choice" -> "incorrect")))
 
-      getElementByCss(view, "#error-message-choice-input").text() must be(messages(choiceError))
+      view must haveGlobalErrorSummary
+      view must haveFieldErrorLink("choice", "#choice")
+
+      view.select("#error-message-choice-input").text() must be(messages(choiceError))
```
ed0906: `view.getElementById(...)`

app/views/components/error_summary.scala.html
MaciejRewera: 👍

PR #128, opened 2019-10-14
CEDS-1675 Change the case of TXM ids to camelcase

PauloPortugal:
This was the feedback from TxM team.


app/services/audit/AuditService.scala
```diff
@@ -151,5 +151,5 @@ object AuditTypes extends Enumeration {
 }
 object EventData extends Enumeration {
   type Data = Value
-  val EORI, MUCR, DUCR, UCR, UCRType, MessageCode, MovementReference, SubmissionResult, Success, Failure = Value
+  val eori, mucr, ducr, ucr, ucrType, messageCode, movementReference, submissionResult, Success, Failure = Value
```
PatrykRudnicki: What about Success and Failure?

PauloPortugal: That is a value not a key. TxM wanted the keys to be `camelCase`

ed0906: I think they are only used as values, not keys in the JSON

PR #129, opened 2019-10-15
CEDS-1809 Change order of screens for departure

PatrykRudnicki:


app/controllers/MovementDetailsController.scala
```diff
@@ -90,12 +90,8 @@ class MovementDetailsController @Inject()(
...
         (formWithErrors: Form[DepartureDetails]) => Future.successful(Left(departureDetailsPage(formWithErrors))),
         validForm =>
-          customsCacheService.cache[DepartureDetails](movementCacheId, formId, validForm).map { cacheMap: CacheMap =>
-            cacheMap.getEntry[GoodsDeparted](GoodsDeparted.formId) match {
-              case Some(goodsDeparted) if (goodsDeparted.departedPlace == AllowedPlaces.outOfTheUk) =>
-                Right(controllers.routes.TransportController.displayPage())
-              case _ => Right(controllers.routes.SummaryController.displayPage())
-            }
+          customsCacheService.cache[DepartureDetails](movementCacheId, formId, validForm).map { _ =>
+            Right(controllers.routes.LocationController.displayPage())
```
bambuchaAdm: Why we would like to have Either ?

test/views/TransportViewSpec.scala
```diff
@@ -16,102 +16,90 @@
...
+import play.api.i18n.MessagesApi
 import play.twirl.api.Html
-import views.spec.ViewSpec
+import utils.Injector
+import views.html.transport
+import views.spec.UnitViewSpec
+import views.spec.UnitViewSpec.instanceOf
 
-class TransportViewSpec extends ViewSpec with TransportMessages with CommonMessages {
+class TransportViewSpec extends UnitViewSpec with TransportMessages with CommonMessages with Injector {
```
bambuchaAdm: Why we would like to have injector here ?

app/controllers/MovementDetailsController.scala
PatrykRudnicki: Just to be consistent with the method signature, I'm not sure why we're returning Either here, but it's old logic, so probably it was necessary

test/views/TransportViewSpec.scala
PatrykRudnicki: Good point, I'll remove it

PR #130, opened 2019-10-15
CEDS-1805 New Location page

PatrykRudnicki:


app/forms/Location.scala
```diff
@@ -16,39 +16,51 @@
...
+    * Country is in two first characters in Location Code
+    */
+  private def validateCountry: String => Boolean =
+    (input: String) => allCountries.exists(_.countryCode == input.take(2))
+
+  /**
+    * Location Type is defined as third character in Location Code
+    */
+  private def validateLocationType: String => Boolean =
+    (input: String) => input.drop(2).headOption.map(_.toString).map(isContainedIn(correctLocationType)).getOrElse(false)
```
bambuchaAdm: substring(2,3)

test/views/LocationViewSpec.scala
```diff
@@ -19,74 +19,59 @@ package views
...
     }
 
-    "display text input for all fields" in {
+    "display input hint" in {
 
-      getElementById(view, "locationType-label").text() mustBe messages(locationType)
-      getElementById(view, "qualifierCode-label").text() mustBe messages(qualifierCode)
-      getElementById(view, "locationCode-label").text() mustBe messages(locationCode)
-      getElementById(view, "country-label").text() mustBe messages(country)
+      view.getElementById("code-hint").text() mustBe messages("location.hint")
```
bambuchaAdm: maybe without messages.

app/forms/Location.scala
```diff
@@ -16,39 +16,51 @@
...
+      )
   )(Location.apply)(Location.unapply)
 
   def form(): Form[Location] = Form(mapping)
 
+  /**
+    * Country is in two first characters in Location Code
+    */
+  private def validateCountry: String => Boolean =
+    (input: String) => allCountries.exists(_.countryCode == input.take(2))
```
bambuchaAdm: Here you would make `take(2)` for every entry in `allCountries` - what do you think about 
```
(input: String) => {
  val code = input.take(2)
  allCountries.exists(_.countryCode == code)
}
```

app/forms/Location.scala
```diff
@@ -16,39 +16,51 @@
...
+
+  /**
+    * Location Type is defined as third character in Location Code
+    */
+  private def validateLocationType: String => Boolean =
```
bambuchaAdm: why def not val ?

app/forms/Location.scala
PatrykRudnicki: Second comment, yeah I agree. The first one no. `substring(2,3)` for one character string will throw and exception. During using `drop` or `take` you will have just empty string, so the second approach is more suitable for this case.

test/views/LocationViewSpec.scala
PatrykRudnicki: I used messages, to be consistent. In this approach you can easily change messages from stub to real one and all tests will pass.

When I'll have enough time for everything I'll copy helper from exports and test it using this.

app/forms/Location.scala
PatrykRudnicki: Yup, you're right, it should be in separate val.

app/forms/Location.scala
PatrykRudnicki: Mistake, I'll change it

app/forms/Location.scala
```diff
@@ -16,37 +16,56 @@
...
+    */
+  private val validateCountry: String => Boolean = (input: String) => {
+    val countryCode = input.take(2)
+    allCountries.exists(_.countryCode == countryCode)
+  }
+
+  /**
+    * Location Type is defined as third character in Location Code
+    */
+  private val validateLocationType: String => Boolean = (input: String) => {
```
bambuchaAdm: What do you think about moving `correctLocationType` here inside definition

PatrykRudnicki: Solved

PR #131, opened 2019-10-15
CEDS-1725 double code translations

MaciejRewera:


conf/messages-notification-codes-cds.en
```diff
@@ -0,0 +1,96 @@
+
```
PatrykRudnicki: Just style, that you have empty first line :D

PR #132, opened 2019-10-16
CEDS-1803 - Change Position of DUCR/MUCR reference Field

SteveSugden:


app/views/consignment_references.scala.html
```diff
@@ -35,25 +35,37 @@
...
+                      label = messages("consignmentReferences.reference.ducrValue"),
+                      labelClass = Some("bold-small")
+                  ))
+              ),
+              RadioOption(
+                  id = "Mucr",
+                  value = Mucr,
+                  messageKey = "consignmentReferences.reference.mucr",
+                  revealComponent = Some(components.input_text(
+                      field = form("mucrValue"),
```
PatrykRudnicki: Hmm, I think that instead of creating models with two values etc, we can in both `revealComponent` put input text that will point to the same form field.

So if we will stay with a model and everything here we can add `form("referenceValue")` and in line 53 the same. Both inputs will be pointing to the same field.

PatrykRudnicki: Is it possible and will it be working correctly? Or you tried it before and it's not working?

PR #136, opened 2019-10-18
CEDS-1838 - sort submissions in descending order (newest first)

SteveSugden:


SteveSugden:
I did the sorting in the view originally.   Could move it back there?

PR #138, opened 2019-10-18
CEDS-1808 Add links to all confirmation pages

PatrykRudnicki:


app/controllers/ChoiceController.scala
```diff
@@ -44,6 +44,20 @@ class ChoiceController @Inject()(
       .map(data => Ok(choicePage(data.fold(form())(form().fill(_)))))
   }
 
+  def startSpecificJourney(choice: String): Action[AnyContent] = authenticate.async { implicit request =>
+    val correctChoice = Choice(choice)
```
bambuchaAdm: typo

PatrykRudnicki: Where?

app/views/disassociate_ducr_confirmation.scala.html
```diff
@@ -28,11 +28,14 @@
...
     )
 
-    @components.paragraph(messages("movement.confirmation.statusInfo"), Some("status-info"))
+    @components.confirmation_status_info()
 
     <h2 id="what-next">@messages("movement.confirmation.whatNext")</h2>
 
-    @components.paragraph(messages(s"consolidation.confirmation.addOrShut"), Some("next-steps"))
+    <div id="next-steps">
+        <a href="@routes.ChoiceController.startSpecificJourney("associateDUCR")">@messages("consolidation.confirmation.addOrShut.part1")</a> @messages("consolidation.confirmation.addOrShut.part2")
```
PatrykRudnicki: Those weird links will be replaced with better approach: https://stackoverflow.com/questions/28587760/html-parameter-passed-to-messages-gets-escaped-in-play-framework/28587899#28587899

app/views/associate_ducr_confirmation.scala.html
```diff
@@ -29,11 +29,14 @@
...
     )
 
-    @components.paragraph(messages("movement.confirmation.statusInfo"), Some("status-info"))
+    @components.confirmation_status_info()
 
     <h2 id="what-next">@messages("movement.confirmation.whatNext")</h2>
 
-    @components.paragraph(messages(s"consolidation.confirmation.addOrShut"), Some("next-steps"))
+    <div id="next-steps">
+        <a href="@routes.ChoiceController.startSpecificJourney("associateDUCR")">@messages("consolidation.confirmation.addOrShut.part1")</a> @messages("consolidation.confirmation.addOrShut.part2")
```
bambuchaAdm: This enforce structure on translator - i think it would be much better if we would use same solution as title.
Like - we have paragraph main translation and HTML links as parameters.
```
paragraph = Enter {}
link.one = Association
```
and then render would looks like this (this is pseudocode - i'm not sure about API 100%)
```
@messages("paragraph", <a href="routes.Controller.show()">@messages("link.one")</a>)
```
You have full control of text name and it position in paragraph though translation files

PatrykRudnicki: Changed approach

app/views/shut_mucr_confirmation.scala.html
```diff
@@ -22,17 +22,22 @@
 
 @()(implicit request: Request[_], flash: Flash, messages: Messages)
 
+@shutMucrLink = {<a href="@routes.ChoiceController.startSpecificJourney("shutMUCR")">@messages("shutMucr.confirmation.nextSteps.shutMucr")</a>}
```
bambuchaAdm: Can we use variable as parametr to reverse routing?

PatrykRudnicki: Yes. I'll do it in the next PR to merge those changes.

PR #142, opened 2019-10-22
CEDS-1812 Disassociate MUCR

SteveSugden:
This id based heavily on yesterdays "Associate MUCR" changes.   There are some areas of overlap and opportunities for code-reuse which we can possibly create a tech debt ticket for?


app/forms/DisassociateUcr.scala
```diff
@@ -0,0 +1,75 @@
...
+object DisassociateKind {
+  case object Mucr extends DisassociateKind("mucr")
+  case object Ducr extends DisassociateKind("ducr")
+
+  private val lookup = PartialFunction[String, DisassociateKind] {
+    case Mucr.formValue => Mucr
+    case Ducr.formValue => Ducr
+  }
+
+  implicit val formatter: Formatter[DisassociateKind] = new Formatter[DisassociateKind] {
```
ed0906: Still not sure what this is needed for

app/views/disassociate_ucr.scala.html
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+        @components.input_radio(
+            field = form("type"),
+            legend = "",
+            inputs = Seq(
+                RadioOption(
+                    keyPrefix = "disassociate.ucr",
+                    option = DisassociateKind.Ducr.formValue,
+                    revealComponent = components.input_text(form("ducr"), "")  // TODO input label
```
ed0906: We can always add a label, but stick a class on so its screen reader only

app/forms/DisassociateUcr.scala
SteveSugden: As noted above - was modelled on Association.  There are now three places where we need to capture either a Mucr or Ducr in basically the same form (radio option with conditional reveal) and I think the form code could be shared.    Especially as this version does not currently cope with neither ucr value being entered.

PR #144, opened 2019-10-22
CEDS-1595 select validation

bambuchaAdm:


app/forms/AssociateUcr.scala
```diff
@@ -48,29 +50,44 @@ object AssociateKind {
...
+          case Ducr =>
+            if (ducr.isEmpty) Invalid("ducr.error.empty")
+            else {
+              if (validDucr(ducr)) Valid else Invalid("ducr.error.format")
+            }
+          case Mucr =>
+            if (mucr.isEmpty) Invalid("mucr.error.empty")
+            else {
+              if (validMucr(mucr)) Valid else Invalid("mucr.error.format")
+            }
```
PatrykRudnicki: What do you think:
```
if (ducr.isEmpty) Invalid("ducr.error.empty")
else if (validDucr(ducr)) Valid
else Invalid("ducr.error.format")
```

The same for Mucr

app/utils/ContextErrorMaping.scala
```diff
@@ -0,0 +1,51 @@
...
+  private def fixErrors(errors: Seq[FormError]): Seq[FormError] =
+    errors.map {
+      case formError if formError.key.nonEmpty =>
+        formError
+      case formError =>
+        origin.mappings.collectFirst {
+          case mappingForTest if mappingForTest.key.nonEmpty && formError.message.startsWith(mappingForTest.key) =>
+            formError.copy(mappingForTest.key)
+        }.getOrElse(formError)
+    }
```
PatrykRudnicki: What this method is doing?

SteveSugden: "Magic" Patryk, magic.

app/forms/AssociateUcr.scala
bambuchaAdm: To me it looks bad - i will try to get something better

app/utils/ContextErrorMaping.scala
bambuchaAdm: Checks if any of global `FormError` message starts with any of `name` used in form and then if start move that global error to field with that name

app/forms/AssociateUcr.scala
bambuchaAdm: Was in two lines - result of formatter

app/utils/ContextErrorMaping.scala
SteveSugden: Basically, it's looking for FormErrors that don't have a key (i.e. they are not linked to a field on the form) and then adds a key based on the first part of the message matching the id of a field on the form.   So if we have an error like this - FormError("", "ducr.error") and the form contains a field with the id "ducr" then the error gets transformed to FormError("ducr", "ducr.error").

This is all done so that when the error is rendered the field in question gets highlighted.

The only slight downside of this implementation that I can see is that it relies on the error messages starting with the id of the field to which they relate.

PatrykRudnicki: Thank you, yeah, understand now

app/forms/AssociateUcr.scala
ed0906: I prefer @PatrykRudnicki

app/utils/ContextErrorMaping.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import play.api.data.{FormError, Mapping}
+import play.api.data.validation.Constraint
+
+case class ContextErrorMapping[T](origin: Mapping[T]) extends Mapping[T] {
+  override val key: String = origin.key
+  override val mappings: Seq[Mapping[_]] = origin.mappings
+  override val constraints: Seq[Constraint[T]] = origin.constraints
+
+  private def fixErrors(errors: Seq[FormError]): Seq[FormError] =
```
ed0906: I think this needs a more descriptive name

app/utils/ContextErrorMaping.scala
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package utils
+
+import play.api.data.{FormError, Mapping}
+import play.api.data.validation.Constraint
+
+case class ContextErrorMapping[T](origin: Mapping[T]) extends Mapping[T] {
```
ed0906: I like the idea of wrapping the `Mapping` to provide extra e.g multi-field validation however I think it needs to be *intuitive* (or it wont be used) and I'm not convinced this is particularly.

I'd prefer if we could design the signature of what we want to do first e.g.

```
val mapping: Mapping[T] = 
Forms.mapping(
    "field1" -> text().verifying("field1.error", isEmpty or isValidField),
    "field2" -> text().verifying("field2.error", isEmpty or isValidField)
)(SomeClass.apply)(SomeClass.unapply)
    .verifying(Seq("field1.error", "field2.error"), someClass => isNonEmpty(someClass.field1) or isNonEmpty(someClass.field2))
```

Then make it work.

e.g. with a wrapper class e.g.
```
implicit class ContextErrorMapping[T](origin: Mapping[T]) extends Mapping[T] {
      def verifying(fieldErrors: Seq[String], condition: T => Boolean) = {
            // Some code to add the errors
      }
}
```

That way it stays intuitive

bambuchaAdm: But here we won't produce error at two field at one. Put in point - API based on convention and there is no straight obvious way to use it. 

I will migrate it to solution by conditional mapping

PR #149, opened 2019-10-24
CEDS-1826 internal UI changes

MaciejRewera:


conf/application.conf
```diff
@@ -90,8 +90,8 @@ microservice {
       submit-movements = "/movements"
       submit-consolidation = "/consolidation"
       fetch-notifications = "/notifications"
-      fetch-all-submissions = "/movements"
-      fetch-single-submission = "/movements"
+      fetch-all-submissions = "/submissions"
+      fetch-single-submission = "/submissions"
```
PatrykRudnicki: I know that it was before, but IMO two vals with the same config doesn't give any value.

PR #150, opened 2019-10-24
CEDS-1595 Add notification handling for MUCR consolidation 

bambuchaAdm:


app/models/viewmodels/notificationspage/NotificationPageSingleElementFactory.scala
```diff
@@ -34,8 +34,8 @@ class NotificationPageSingleElementFactory @Inject()(responseConverterProvider:
 
   def build(submission: SubmissionFrontendModel)(implicit messages: Messages): NotificationsPageSingleElement =
     submission.actionType match {
-      case Arrival | Departure | DucrDisassociation | ShutMucr => buildForRequest(submission)
-      case DucrAssociation                                     => buildForDucrAssociation(submission)
+      case Arrival | Departure | DucrDisassociation | MucrAssociation | MucrDisassociation | ShutMucr => buildForRequest(submission)
```
SteveSugden: Should the MucrAssociation not be treated the same as the DucrAssociation?

bambuchaAdm: DUCR has multiple value - MUCR could have only one single

PR #159, opened 2019-10-29
CEDS-1875 Fix formatting with double digits

bambuchaAdm:
Use Instant format for communication to backend
Use double digit formatting on form level
Use single error for invalid data

### DIFF
--- a/app/forms/common/Date.scala
+++ b/app/forms/common/Date.scala
@@ -59,7 +59,8 @@ object Date {
         d <- day
         m <- month
         y <- year
-      } yield Try(LocalDate.of(y, m, d))
+        date <- Try(LocalDate.of(y, m, d))
+      } yield date
 
     def validate(day: Try[Int], month: Try[Int], year: Try[Int]): Boolean = build(day, month, year).isSuccess
 
--- a/app/forms/common/Date.scala
+++ b/app/forms/common/Date.scala
@@ -59,8 +59,7 @@ object Date {
         d <- day
         m <- month
         y <- year
-        date <- Try(LocalDate.of(y, m, d))
-      } yield date
+      } yield LocalDate.of(y, m, d)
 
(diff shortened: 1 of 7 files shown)


app/forms/common/Date.scala
```diff
@@ -51,39 +52,41 @@ object Date {
...
-        case (Some(d), Some(m), Some(y)) => Try(LocalDate.of(y, m, d)).isSuccess
-        case _                           => false
-      }
+    def build(day: Try[Int], month: Try[Int], year: Try[Int]): Try[LocalDate] =
+      for {
+        d <- day
+        m <- month
+        y <- year
+        date <- Try(LocalDate.of(y, m, d))
+      } yield date
```
PatrykRudnicki: Can we move Try to yield?
```
for {
  d <- day
  m <- month
  y <- year
} yield Try(LocalDate.of(y, m, d)
```

app/forms/common/Time.scala
```diff
@@ -36,26 +37,30 @@ object Time {
...
   val minuteKey = "minute"
 
-  private val correctHour: String => Boolean = (hour: String) => Try(hour.toInt).map(value => value >= 0 && value <= 23).getOrElse(false)
+  val mapping: Mapping[Time] = {
+    def build(hour: Try[Int], minutes: Try[Int]): Try[LocalTime] =
+      for {
+        h <- hour
+        m <- minutes
+        time <- Try(LocalTime.of(h, m))
+      } yield time
```
PatrykRudnicki: The same here, Try to yield

app/forms/common/Date.scala
bambuchaAdm: No i could not this would lead to `Try[Try[Int]]` optionally i can remove try and LocalDate but i'am not sure how it would handle exceptions

PR #163, opened 2019-11-06
CEDS-1860 Improve algorithm in ViewSubmissionsController

MaciejRewera:
The algorithm is for matching Submissions with Notifications by their
conversation ID.

The time complexity used to be s * n, where:
 - s - number of Submissions
 - n - number of Notifications
For large amounts of those it would get slower quite quickly.
Additionally it was sorting Notifications before the process which was
unnecessary as they were being filtered later.

The new solution makes use of groupBy method and therefore comes down
to logarithmic complexity.

### DIFF
--- a/app/controllers/MovementsController.scala
+++ b/app/controllers/MovementsController.scala
@@ -28,7 +28,7 @@ import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import views.html.movements
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 
 class MovementsController @Inject()(
   authenticate: AuthAction,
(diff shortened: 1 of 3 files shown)


app/controllers/MovementsController.scala
```diff
@@ -38,18 +38,30 @@ class MovementsController @Inject()(
...
-    } yield Ok(movementsPage(sort(submissionsWithNotifications)))
+  private def matchNotificationsAgainstSubmissions(
+    submissions: Seq[SubmissionFrontendModel],
+    notifications: Seq[NotificationFrontendModel]
+  ): Seq[(SubmissionFrontendModel, Seq[NotificationFrontendModel])] = {
+    val groupedNotifications: Map[String, Seq[NotificationFrontendModel]] = notifications.groupBy(_.conversationId).withDefaultValue(Seq.empty)
+
+    submissions.map { submission =>
+      (submission, groupedNotifications(submission.conversationId))
+    }
```
PatrykRudnicki: I think it'll be enough to use `zip` method from `Seq`

https://www.scala-lang.org/api/2.12.3/scala/collection/Seq.html#zip[B](that:scala.collection.GenIterable[B]):Seq[(A,B)]

MaciejRewera: No, it is not. I thought about it, but no. Once notifications are grouped, they're in a Map and they are not ordered to match submissions.
You could `zip` them, but you would have different notifications assigned to submissions every time there is a new notification.

Would be fun though...

PR #168, opened 2019-11-18
CEDS-1958 Align Caching in Movements & Internal Service

ed0906:
- Removes Keystore from Movements & uses a similar model to the Internal UI
- Uses the JourneyRefiner used in the Internal UI which allows us to have less controller validation


app/controllers/actions/JourneyRefiner.scala
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+  override protected def refine[A](request: AuthenticatedRequest[A]): Future[Either[Result, JourneyRequest[A]]] =
+    refiner(request, Seq.empty[JourneyType]: _*)
+
+  def apply(types: JourneyType*): ActionRefiner[AuthenticatedRequest, JourneyRequest] = new ActionRefiner[AuthenticatedRequest, JourneyRequest] {
+    override protected def refine[A](request: AuthenticatedRequest[A]): Future[Either[Result, JourneyRequest[A]]] =
+      refiner(request, types: _*)
+
+    override protected def executionContext: ExecutionContext = exc
```
PatrykRudnicki: hmm, why you need this? You can override thing from implicit. You just need to change name from `exc` to `executionContext`

PatrykRudnicki: And line 32 contains this: `override protected def executionContext: ExecutionContext = exc` so you're overriding this two times

ed0906: It's an anonymous inner class

ed0906: Ah its also a `def` not a val

app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -34,79 +34,58 @@ import scala.util.{Failure, Success}
...
-    postRequest(Departure, movementRequest)
-
-  private def postRequest(
-    actionType: ActionType,
-    movementRequest: MovementRequest
-  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] =
+  def submit(request: MovementRequest)(implicit hc: HeaderCarrier): Future[Unit] =
     httpClient
-      .POST[MovementRequest, HttpResponse](movementSubmissionUrl(actionType), movementRequest, JsonHeaders)
+      .POST[MovementRequest, HttpResponse](appConfig.customsDeclareExportsMovements + Movements, request, JsonHeaders)
```
bambuchaAdm: are we have to do this concatenation for each request ?

app/controllers/actions/JourneyRefiner.scala
bambuchaAdm: it's dev due to on inline and initialization problems - could be override as val

PR #169, opened 2019-11-20
CEDS-1390 Use Continue instead Save and Continue

bambuchaAdm:
Refactor submit button to take messages outside

Little reformatting templates where needed

### DIFF
--- a/app/views/arrival_details.scala.html
+++ b/app/views/arrival_details.scala.html
@@ -46,6 +46,8 @@
             hint = Some(messages("arrivalDetails.time.hint"))
         )
 
-        @components.submit_button()
+        @components.submit_button {
+            @messages("site.continue")
+        }
     }
 }
--- a/app/views/arrival_details.scala.html
+++ b/app/views/arrival_details.scala.html
@@ -46,8 +46,6 @@
             hint = Some(messages("arrivalDetails.time.hint"))
         )
 
-        @components.submit_button {
-            @messages("site.continue")
-        }
+        @components.submit_button(messages("site.continue"))
     }
 }
--- a/app/views/components/submit_button.scala.html
+++ b/app/views/components/submit_button.scala.html
@@ -14,8 +14,8 @@
  * limitations under the License.
  *@
 
-@(messageKey: String = "site.save_and_continue")(implicit messages: Messages)
+@(message: Html)
 
 <div class="section">
-    <button id="submit" class="button">@messages(messageKey)</button>
+    <button id="submit" class="button">@message</button>
 </div>
--- a/app/views/components/submit_button.scala.html
+++ b/app/views/components/submit_button.scala.html
@@ -14,7 +14,7 @@
  * limitations under the License.
  *@
 
-@(message: Html)
+@(message: String)
 
 <div class="section">
     <button id="submit" class="button">@message</button>
--- a/app/views/main_template.scala.html
+++ b/app/views/main_template.scala.html
@@ -21,29 +21,29 @@
 @this(govuk_wrapper: views.html.govuk_wrapper, sidebar_di: Sidebar, article_di: Article)
 
 @(title: Title,
-        sidebarLinks: Option[Html] = None,
-        contentHeader: Option[Html] = None,
-        bodyClasses: Option[String] = None,
-        mainClass: Option[String] = None,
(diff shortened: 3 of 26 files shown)


app/views/arrival_details.scala.html
```diff
@@ -46,6 +46,8 @@
             hint = Some(messages("arrivalDetails.time.hint"))
         )
 
-        @components.submit_button()
+        @components.submit_button {
+            @messages("site.continue")
+        }
```
PatrykRudnicki: Is it worth to split it to two lines? I think one will be better. Just style comment

app/views/components/submit_button.scala.html
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  *@
 
-@(messageKey: String = "site.save_and_continue")(implicit messages: Messages)
+@(message: Html)
```
PatrykRudnicki: Why a messages is `Html` not `String`?

app/views/main_template.scala.html
```diff
@@ -21,29 +21,29 @@
...
-  bodyClasses: Option[String] = None,
-  mainClass: Option[String] = None,
-  scriptElem: Option[Html] = None,
-  fullWidth: Boolean = false)(mainContent: Html)(implicit request : Request[_], messages: Messages)
+        sidebarLinks: Option[Html] = None,
+        contentHeader: Option[Html] = None,
+        bodyClasses: Option[String] = None,
+        mainClass: Option[String] = None,
+        scriptElem: Option[Html] = None,
+        fullWidth: Boolean = false)(mainContent: Html)(implicit request: Request[_], messages: Messages)
```
PatrykRudnicki: I'm not sure that this is the expected format. arguments should be align in one line, now title is in one place rest in the different

app/views/main_template.scala.html
```diff
@@ -21,29 +21,29 @@
...
-@govuk_wrapper(title = title.title,
-               mainClass = mainClass,
-               bodyClasses = bodyClasses,
-               sidebar = sidebar,
-               contentHeader = contentHeader,
-               mainContent = article_di(mainContent, articleClasses = if(fullWidth) Some("full-width") else None),
-               serviceInfoContent = serviceInfoContent,
-               scriptElem = scriptElem
-)
+    @govuk_wrapper(title = title.title,
```
PatrykRudnicki: Similar here, it's not inside something, it's separate element, should start from the left

app/views/components/submit_button.scala.html
ed0906: Agreed, I see no need for this to be a `Html`

ed0906: We could just do this as
```
@(text: String = "")(implicit messages: Messages)
<div class="section">
   <button id="submit" class="button">
        @if(text.nonEmpty) text else @messages("site.save_and_continue")) 
   </button>
</div>
```

app/views/arrival_details.scala.html
bambuchaAdm: Could be - on the other hand if treated as HTML messages outcome is inside

app/views/components/submit_button.scala.html
bambuchaAdm: Still passing messages - this makes controller as dummy as possible without any logic inside

SteveSugden: Agreed.  Have already suggested that.     I think this whole PR is overly complicated and that components that effectively render "static" text should be free to do their own message look-up.

The rule about supplying components with message strings rather than keys should only apply where components render different messages.

bambuchaAdm: Moreover I think it is not responsibility of component to provide to itself content

PatrykRudnicki: IMO there shouldn't be a default value for it. It's more common for us to use Continue etc not always Save and continue. I'll change it to String, but without any dependency with messages and without default value

app/views/main_template.scala.html
PauloPortugal: I agree with @PatrykRudnicki

app/views/main_template.scala.html
PauloPortugal: I agree with Patry

PR #171, opened 2019-11-27
CEDS-1783 Add visually hidden text to Change buttons

PauloPortugal:


test/views/ShutMucrSummaryViewSpec.scala
```diff
@@ -65,7 +65,7 @@ class ShutMucrSummaryViewSpec extends UnitViewSpec with CommonMessages {
       val changeButton = view.getElementById("shutMucr-change")
 
       changeButton must haveHref(routes.ShutMucrController.displayPage())
-      changeButton.text() mustBe messages("site.edit")
+      changeButton.text() contains messages("site.edit")
```
PatrykRudnicki: Are you sure about `contains`?

bambuchaAdm: maybe better include - there was quite different behaves in scalatest about those matchers

PauloPortugal: yes good point. Done.

PR #177, opened 2019-12-05
CEDS-1783 Amend associate MUCR change tag id

PauloPortugal:
* I missed an updated of the mucr change tag id from `mucr-change` to
  `associate_mucr-change`. This change is to align the ids with the same
  format in other places.

* This change was already done in the acceptance tests suite and the
  movements-regression tests failed.

### DIFF
--- a/app/views/associate_ucr_summary.scala.html
+++ b/app/views/associate_ucr_summary.scala.html
@@ -59,7 +59,7 @@
                     <td id="mucr-type">@messages("associate.ucr.summary.kind.mucr")</td>
                     <td id="mucr-reference">@mucr</td>
                     <td class="text-right">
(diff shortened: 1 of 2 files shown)


app/views/associate_ucr_summary.scala.html
```diff
@@ -59,7 +59,7 @@ <h2 id="master-consignment">@messages("associate.ucr.summary.masterConsignment")
                     <td id="mucr-type">@messages("associate.ucr.summary.kind.mucr")</td>
                     <td id="mucr-reference">@mucr</td>
```
MaciejRewera: I am aware it is not a part of this change, but should these elements have also their id's changed?

PauloPortugal: I am going to make a note of that when we move to the GDS design system! Hawk Eye

PR #178, opened 2019-12-05
CEDS-1958 Add Eori to audit events

ed0906:


app/services/audit/AuditService.scala
```diff
@@ -22,7 +22,7 @@ import javax.inject.Named
 import models.cache.{Answers, ArrivalAnswers, DepartureAnswers, JourneyType}
 import models.requests.MovementRequest
 import play.api.Logger
-import play.api.libs.json.{JsObject, Json}
+import play.api.libs.json.{JsObject, JsString, JsValue, Json}
```
PatrykRudnicki: You're not using `JsValue`

PR #180, opened 2019-12-06
CEDS-2022 Allow Association by MUCR

ed0906:


ed0906:
> Do we need to add backend changes to handle this type?

Looks like they were already done a while ago

PR #184, opened 2019-12-09
CEDS-2024 Introduce govuk styles dependencies

MaciejRewera:


app/assets/stylesheets/customsdecexfrontend-app.scss
```diff
@@ -9,15 +9,15 @@
...
 @import 'partials/_lists.scss';
 @import 'partials/_layout.scss';
 
 .text-right {
   text-align: right;
-}
\ No newline at end of file
+}
+
+$govuk-assets-path: '/customs-movements/assets/';
```
bambuchaAdm: I am not sure that variable definition should goes before import

PauloPortugal: I don't think it matters, we could try this approach

app/views/components/errorSummary.scala.html
```diff
@@ -0,0 +1,36 @@
...
+        content = Text(messages(error.message, error.args: _*))
+    )
+}}
+
+@if(errorList.nonEmpty) {
+    @govukErrorSummary(
+        ErrorSummary(
+            errorList = errorList,
+            title = Text(messages("error.summary.title"))
+        ))
```
PatrykRudnicki: Just style, close the brackets in column that you open it, more readable

So last bracket to the new line. Style comment

PR #186, opened 2019-12-11
CEDS-2024 Integrate GDS layout into Choice page

PauloPortugal:


app/views/choice_page.scala.html
```diff
@@ -14,41 +14,78 @@
...
+                    isPageHeading = true,
+                    classes = "govuk-fieldset__legend--xl"
+                ))
+            )),
+            items = Seq(
+                RadioItem(
+                    id = Some(Arrival.toString.toLowerCase),
+                    value = Some(Arrival.value.toString),
+                    content = Text(messages(s"movement.choice.${Arrival.toString.toLowerCase}.label")),
+                    checked = form("choice").value.contains(Arrival.toString.toLowerCase)
```
bambuchaAdm: why we match against field not form bind value itself

test/unit/helpers/views/ChoiceMessages.scala
```diff
@@ -22,9 +22,9 @@ trait ChoiceMessages {
 
   val arrivalDecLabel: String = "movement.choice.arrival.label"
   val departureDecLabel: String = "movement.choice.departure.label"
-  val associateDecLabel: String = "movement.choice.associate.label"
-  val disassociateDecLabel: String = "movement.choice.disassociateDucr.label"
-  val shutMucrLabel: String = "movement.choice.shutMucr.label"
+  val associateDecLabel: String = "movement.choice.associateucr.label"
```
bambuchaAdm: Maybe would be worth to drop this helper trait?

test/unit/views/ChoiceViewSpec.scala
```diff
@@ -85,12 +85,12 @@ class ChoiceViewSpec extends UnitViewSpec with ChoiceMessages with CommonMessage
 
       val view = createView(Choice.form())
 
-      view.select("div.multiple-choice:nth-child(2)").text() must be(messages(arrivalDecLabel))
-      view.select("div.multiple-choice:nth-child(3)").text() must be(messages(associateDecLabel))
-      view.select("div.multiple-choice:nth-child(4)").text() must be(messages(disassociateDecLabel))
-      view.select("div.multiple-choice:nth-child(5)").text() must be(messages(shutMucrLabel))
-      view.select("div.multiple-choice:nth-child(6)").text() must be(messages(departureDecLabel))
-      view.select("div.multiple-choice:nth-child(7)").text() must be(messages(viewSubmissionsLabel))
+      view.getElementsByClass("govuk-label govuk-radios__label").get(0).text() must be(messages(arrivalDecLabel))
```
bambuchaAdm: Maybe would be worth to pick little different class - we could go via utility way like `[for='id-for-field']` or go little by BEM
```.govuk-radios__label.cds_choice__arrival``` - what i'am not like is we depend on order here.

File: test/unit/views/ChoiceViewSpec.scala [hunk 4]
```diff
@@ -179,8 +179,8 @@ class ChoiceViewSpec extends UnitViewSpec with ChoiceMessages with CommonMessage
   }
 
   private def ensureRadioIsChecked(view: Document, elementId: String): Unit = {
-    val option = view.getElementById(elementId)
-    option.attr("checked") mustBe "checked"
+    val option = view.getElementById(elementId).getElementsByAttribute("checked")
+    option.size() mustBe 1
```
bambuchaAdm: Now we are not sure that this is expected element

PauloPortugal: We are because we use the `elementId`

test/unit/helpers/views/ChoiceMessages.scala
PauloPortugal: YESSSSSS!!! Thanks! Done.

test/unit/views/ChoiceViewSpec.scala
PauloPortugal: Good recommendation.

app/views/choice_page.scala.html
PauloPortugal: Awesome! :) Thank you!

app/views/choice_page.scala.html
```diff
@@ -14,41 +14,78 @@
...
+                    checked = form("choice").value.contains(DisassociateUCR.value.toString)
+                ),
+                RadioItem(
+                    id = Some("shut_mucr"),
+                    value = Some(ShutMUCR.value.toString),
+                    content = Text(messages(s"movement.choice.${ShutMUCR.toString.toLowerCase}.label")),
+                    checked = form("choice").value.contains(ShutMUCR.value.toString)
+                ),
+                RadioItem(
+                    id = Some(Departure.toString.toLowerCase()),
```
MaciejRewera: Why do you use `Departure.toString` here but above you have hard-coded Strings?

PauloPortugal: I was trying to avoid hardcoding all values, but probably it is better to be consistent

PR #188, opened 2019-12-11
CEDS 2024 choice fix

SteveSugden:
Fix required for links in error summary to work

### DIFF
--- a/test/unit/views/ChoiceViewSpec.scala
+++ b/test/unit/views/ChoiceViewSpec.scala
@@ -82,26 +82,26 @@ class ChoiceViewSpec extends UnitViewSpec with CommonMessages with Injector {
       backButton.attr("href") must be(controllers.routes.StartController.displayStartPage().url)
     }
 
-    "display 4 radio buttons with labels" in {
+    "display 6 radio buttons with labels" in {
 
       val view = createView(Choice.form())
 
-      view.getElementsByAttributeValue("for", "arrival").text() must be(messages("movement.choice.arrival.label"))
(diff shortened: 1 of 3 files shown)


test/unit/views/ChoiceViewSpec.scala
```diff
@@ -82,26 +82,26 @@ class ChoiceViewSpec extends UnitViewSpec with CommonMessages with Injector {
...
 
       val view = createView(Choice.form())
 
-      view.getElementsByAttributeValue("for", "arrival").text() must be(messages("movement.choice.arrival.label"))
-      view.getElementsByAttributeValue("for", "departure").text() must be(messages("movement.choice.departure.label"))
-      view.getElementsByAttributeValue("for", "submissions").text() must be(messages("movement.choice.submissions.label"))
-      view.getElementsByAttributeValue("for", "associate").text() must be(messages("movement.choice.associateucr.label"))
-      view.getElementsByAttributeValue("for", "disassociate").text() must be(messages("movement.choice.disassociateucr.label"))
-      view.getElementsByAttributeValue("for", "shut_mucr").text() must be(messages("movement.choice.shutmucr.label"))
+      view.getElementsByAttributeValue("for", "choice").text() must be(messages("movement.choice.arrival.label"))
```
PatrykRudnicki: I think it'll be good to have `choice-1` to have the same pattern for all options.

SteveSugden: The ids are generated by the GDS component.  We need the first radio option to have an id of "choice" because that is the name of the field on the form and therefore when there is a validation error with that field the link on the error summary has a href of "#choice" (ie links to the first radio option)

PR #190, opened 2019-12-11
GDS design for mucr-options page

janeturner:


app/views/gds_main_template.scala.html
```diff
@@ -27,7 +27,7 @@
         phaseBanner: components.phaseBanner
 )
 
-@(  title: Title,
+@(  title: String,
```
SteveSugden: Changes to this template aren't needed.  The bug has been fixed.

app/views/mucr_options.scala.html
```diff
@@ -16,42 +16,72 @@
...
+formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF,
+appConfig: AppConfig)
 
 @(form: Form[MucrOptions])(implicit request: Request[_], messages: Messages)
 
-@main_template(title = Title("mucrOptions.title", None)) {
-    @helper.form(consolidations.routes.MucrOptionsController.save(), 'autoComplete -> "off") {
-        @helper.CSRF.formField
+@govukLayout(
+    title = messages("mucrOptions.title"),
```
SteveSugden: You can go back to `Title("mucrOptions.title", None))` here

app/views/mucr_options.scala.html
```diff
@@ -16,42 +16,72 @@
...
+            )),
+            items = Seq(
+                RadioItem(
+                    value = Some("create"),
+                    content = Text(messages("mucrOptions.create")),
+                    checked = form("createOrAdd").value.contains(),
+                    conditionalHtml = Some(govukInput(Input(
+                                        id = "mucrOptions.create",
+                                        name = "mucrOptions.create",
+                                        classes = "govuk-!-width-one-third",
```
SteveSugden: Why does this RadioItem have a label and the other one doesn't?   If you wait for Maciej to merge in his latest PR he has created a re-usable component to use for the Input field.

app/views/gds_main_template.scala.html
MaciejRewera: This change require all pages using this component to be updated. Please go through all pages that are using it and make sure the title is displayed correctly.
As a side note: if tests have not caught thi, it might be that they're using `stubMessages` from Play! helpers. This is not what we really want, because this stub has no mapping for any key, so it always returns the key itself. You may end up displaying keys only and the tests won't tell you that.
What you want to use is `MessagesStub` trait.

app/views/mucr_options.scala.html
```diff
@@ -16,42 +16,72 @@
...
+                                        label = Label(
+                                            content = Text(messages("mucrOptions.create.reference"))
+                                        ),
+                                        attributes = Map("spellcheck" -> "false")
```
MaciejRewera: Nit: spellcheck attribute is obsolete. We've never been using it anywhere.

test/unit/views/MucrOptionsViewSpec.scala
```diff
@@ -16,35 +16,33 @@
...
 import forms.MucrOptions
 import helpers.views.CommonMessages
 import org.jsoup.nodes.Document
 import play.api.data.Form
-import views.spec.UnitViewSpec
+import views.html.mucr_options
+import views.spec.{UnitViewSpec, ViewMatchers}
 
-class MucrOptionsViewSpec extends UnitViewSpec with CommonMessages {
+class MucrOptionsViewSpec extends UnitViewSpec with CommonMessages with ViewMatchers with Injector {
```
MaciejRewera: This will need more discussion inside dev team, but there is also `ViewSpec`. It is better to use it as it inherits from `MessagesStub` mentioned earlier.

app/views/mucr_options.scala.html
```diff
@@ -16,42 +16,72 @@
...
+    title = messages("mucrOptions.title"),
+    backButtonTitle = messages("site.back"),
+    backButtonUrl = controllers.routes.ChoiceController.displayChoiceForm().url) {
 
-        @components.back_link(routes.ChoiceController.displayChoiceForm())
+    @formHelper(action = consolidations.routes.MucrOptionsController.save(), 'autoComplete -> "off") {
+        @errorSummary(form.errors)
 
-        @components.error_summary(form.errors)
+        <span class="govuk-caption-xl">@messages("associate.heading", "")</span>
```
PatrykRudnicki: Why there is empty String passing to the messages?

If the message patter doesn't suit for some specific situation I'll create a new message for this page. In general reusing messages is not a good practice, because if you change something you'll change in few places. I think every page should have his own messages keys

test/unit/views/MucrOptionsViewSpec.scala
MaciejRewera: I cannot findn any testst for errors here. Could you please add them?

app/views/gds_main_template.scala.html
SteveSugden: Ignore Maciej (just this once) - "This change require all pages using this component to be updated. Please go through all pages that are using it and make sure the title is displayed correctl" .

test/unit/views/MucrOptionsViewSpec.scala
SteveSugden: While we're having a discussion on message validation, @ed0906 created a "AllMessageKeysAreMandatoryMessages" in the internal service (which appears to not be "switched on" in movements) that will result in view specs failing if they render a page that tries to look up a message key that does not exist.

Should we try to use that in movements?   It saves having to write specific tests to assert that a message key is defined.

app/views/mucr_options.scala.html
SteveSugden: done

app/views/mucr_options.scala.html
SteveSugden: done

test/unit/views/MucrOptionsViewSpec.scala
SteveSugden: Updated to use ViewSpec & added tests for errors

test/unit/views/MucrOptionsViewSpec.scala
```diff
@@ -16,50 +16,48 @@
...
-
-    "have the correct label for create new" in {
-      view.getElementById("mucrOptions.create-label").text() mustBe "mucrOptions.create"
+      page(MucrOptions.form).getElementById("section-header") must containMessage("mucrOptions.heading")
     }
 
-    "have the correct label for add to existing" in {
-      view.getElementById("mucrOptions.add-label").text() mustBe "mucrOptions.add"
-    }
+    // TODO - re-enable when labels added back to input fields
```
PauloPortugal: :+1:

PR #191, opened 2019-12-12
CEDS-2024 refactor Title component

SteveSugden:
- Refactored the Title component to try and prevent incorrect use.  Aim is to ensure that the page title is always rendered as <title> - <optional section> - <service name> - GOV.UK

- Refactored associate_ucr to use the new exportsInputText component

- Created new sectionHeader component for use with GDS pages

### DIFF
--- a/app/views/arrival_details.scala.html
+++ b/app/views/arrival_details.scala.html
@@ -22,7 +22,7 @@
 
 @(form: Form[ArrivalDetails])(implicit request: Request[_], messages: Messages)
 
-@main_template(title = Title("arrivalDetails.header", None)) {
+@main_template(title = Title("arrivalDetails.header")) {
     @helper.form(routes.MovementDetailsController.saveMovementDetails(), 'autoComplete -> "off") {
         @helper.CSRF.formField
 
--- a/app/views/arrival_reference.scala.html
+++ b/app/views/arrival_reference.scala.html
@@ -21,7 +21,7 @@
 
 @(form: Form[ArrivalReference])(implicit request: Request[_], messages: Messages)
 
-@main_template(Title("arrivalReference.question", None)) {
+@main_template(Title("arrivalReference.question")) {
 
     @helper.form(controllers.routes.ArrivalReferenceController.submit(), 'autoComplete -> "off") {
         @helper.CSRF.formField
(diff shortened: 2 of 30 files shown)


app/views/Title.scala
```diff
@@ -18,13 +18,16 @@ package views
 
 import play.api.i18n.Messages
 
-case class Title(title: String)
+case class Title(headingKey: String, sectionKey: Option[String] = None, headingArgs: Seq[String] = Seq.empty) {
```
PatrykRudnicki: I'll vote for not passing keys to the components, but specific values

app/views/Title.scala
```diff
@@ -18,13 +18,16 @@ package views
...
-    if (sectionKey.isEmpty) {
-      Title(messages("title.format", messages(headingKey), messages("service.name")))
-    } else {
-      Title(messages("title.withSection.format", messages(headingKey), messages(sectionKey.getOrElse("")), messages("service.name")))
+  def format(implicit messages: Messages): String = {
+    val heading = messages(headingKey, headingArgs: _*)
+    val service = messages("service.name")
+
+    sectionKey match {
+      case Some(section) => messages("title.withSection.format", heading, messages(section), service)
```
PatrykRudnicki: Nested messages, messages and inside heading with messages, another messages, and service that contains messages. If we will cover it by tests which will check autmatically if the keys exists it will fail

PatrykRudnicki: I'm not sure that in that case it'll make some problems, but we had a problems with nested messages

app/views/Title.scala
```diff
@@ -18,13 +18,16 @@ package views
 
 import play.api.i18n.Messages
 
-case class Title(title: String)
```
SteveSugden: The advantage of getting rid of the 'string' constructor and apply method is that if the Title is (incorrectly) constructed with a translated message (e.g. `Title(messages("consignmentReferences.title"))`) it will still render the title correctly with the formatted "title - service - GOV.UK"

app/views/Title.scala
SteveSugden: So you're suggesting that we construct the Title with the translated message parts (i.e. heading and optional sub-heading)?  And get the Title to use those values as inputs to "title.withSection.format" etc?

I can do that if we get another vote for it

PatrykRudnicki: You're right, it was a different case

PR #193, opened 2019-12-12
CEDS-2024 Add Label to all input text fields

PauloPortugal:


app/views/associate_ucr.scala.html
```diff
@@ -26,6 +26,7 @@
         govukButton: GovukButton,
         govukRadios: GovukRadios,
         govukInput: GovukInput,
+        exportsInputText: components.gds.exportsInputText,
         errorSummary: components.errorSummary,
         formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF)
```
PatrykRudnicki: What do you think about importing those elements instead of putting whole path here?

app/views/associate_ucr.scala.html
```diff
@@ -58,27 +58,17 @@
...
-
-@conditionalHtml(field: String) = {
-    @govukInput(Input(
-        id = field,
-        name = field,
-        value = form(field).value,
-        errorMessage = form(field).error.map(err => ErrorMessage(content = Text(messages(err.message, err.args:_*)))),
-        classes = "govuk-!-width-two-thirds"
-    ))
-}
```
MaciejRewera: 👍 🎉

app/views/choice_page.scala.html
```diff
@@ -17,15 +17,13 @@
...
 @import forms.Choice
-@import config.AppConfig
 @import views.Title
 
 @this(  govukLayout: views.html.gds_main_template,
 govukButton: GovukButton,
 govukRadios: GovukRadios,
 errorSummary: components.errorSummary,
-formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF,
-appConfig: AppConfig)
```
MaciejRewera: 👍

app/views/disassociate_ucr.scala.html
```diff
@@ -54,13 +53,13 @@
                     value = Some(DisassociateKind.Ducr.formValue),
                     content = Text(messages("disassociate.ucr.ducr")),
                     checked = form("kind").value.contains(DisassociateKind.Ducr.formValue),
-                    conditionalHtml = Some(exportsInputText(form("ducr")))
+                    conditionalHtml = Some(exportsInputText(form("ducr"), "mucrOptions.ducr.label"))
```
SteveSugden: Should these be using the same message key for all forms?  If they are really supposed to be the same text then maybe some global/site labels?

test/unit/views/DisassociateUcrViewSpec.scala
```diff
@@ -96,6 +96,12 @@ class DisassociateUcrViewSpec extends ViewSpec with Injector {
       }
     }
 
+    "form contains 'MUCR' with value" in {
+      val mucrView = createView(DisassociateUcr.form.fill(DisassociateUcr(Mucr, ducr = None, mucr = Some("1234"))))
+      mucrView.getElementsByAttributeValue("for", "mucr").first() must containMessage("mucrOptions.mucr.label")
+      mucrView.getElementsByAttributeValue("for", "ducr").first() must containMessage("mucrOptions.ducr.label")
+    }
+
```
MaciejRewera: Should there be an update to the following tests as well? ("form contains 'DUCR' with value")

app/views/associate_ucr.scala.html
PauloPortugal:
```
@import components.gds._
@import components._
```
?

PatrykRudnicki: Yup, and `uk.gov.hmrc.play.views.html.helpers._`

Or specific elements like `@import components.errorSummary`, but not sure if in this case there is no warning

app/views/disassociate_ucr.scala.html
PauloPortugal: Good point.
I have gone with 
```
site.inputText.ducr.label = Enter an existing DUCR
site.inputText.mucr.label = Enter an existing MUCR
site.inputText.newMucr.label = Enter a new MUCR
site.inputText.newMucr.label.hint = For example ‘GB/123456789100-AB123’
```

app/views/associate_ucr.scala.html
PauloPortugal: Done.

test/unit/views/DisassociateUcrViewSpec.scala
PauloPortugal: Done.

PR #202, opened 2019-12-16
CEDS-2027 movement details

MaciejRewera:


app/views/components/gds/inputDate.scala.html
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+@import forms.common.Date.{dayKey, monthKey, yearKey}
+
+@this(dateInput: GovukDateInput)
+
+@(
+  field: Field,
```
ed0906: Looks like you just use this to get `field.name` maybe just pass that in instead?

app/views/components/gds/inputTime.scala.html
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+@import forms.common.Time.{hourKey, minuteKey}
+
+@this(dateInput: GovukDateInput)
+
+@(
+  field: Field,
```
ed0906: Similar, looks like you just use this to get `field.name`. Maybe just pass that in instead?

MaciejRewera:
I've addressed you comment. This was mostly copied from the previous implementation. But it turned out to be better after changed proposed.
Had to force-push as I've removed a commit from thte history.

MaciejRewera:
Another force-push, as I did not rebase.

PR #203, opened 2019-12-17
Ceds 2027 transport

iservport:
Switch movements service single text input pages to the gov design system - transport page.


app/views/transport.scala.html
```diff
@@ -15,56 +15,126 @@
...
+            errorMessage =
+              form("transportId").error.map(err => ErrorMessage(content = Text(messages(err.message, err.args:_*)))
+            )
+        ))
 
-        @components.submit_button(messages("site.continue"))
+        @govukInput(Input(
+            id = "nationality",
+            name = "nationality",
+            classes = "govuk-input--width-2",
```
SteveSugden: Should the two input fields not have the same class(Style)?

test/unit/helpers/views/TransportMessages.scala
```diff
@@ -19,6 +19,7 @@ package helpers.views
 trait TransportMessages {
 
   val title = "transport.title"
+  val heading = "transport.heading"
```
SteveSugden: We are tending to get rid of / not use these message helper classes - just reference the message keys directly in the test (which you appear to have done)

test/unit/helpers/views/TransportMessages.scala
```diff
@@ -19,6 +19,7 @@ package helpers.views
 trait TransportMessages {
```
PatrykRudnicki: Just FYI in general we're trying to get rid of traits with messages. It's easier to inline it and use in specific places or create those variables in specific test where we use it

Specially if you're using this message inlined in the test :)

app/controllers/TransportController.scala
```diff
@@ -41,14 +42,18 @@ class TransportController @Inject()(
...
 
   def saveTransport(): Action[AnyContent] = (authenticate andThen getJourney(JourneyType.DEPART)).async { implicit request =>
+    val answers = request.answersAs[DepartureAnswers]
+    val consignmentReference = answers.consignmentReferences.map(_.referenceValue)
     form
       .bindFromRequest()
       .fold(
-        (formWithErrors: Form[Transport]) => Future.successful(BadRequest(transportPage(formWithErrors))),
+        (formWithErrors: Form[Transport]) => Future.successful(BadRequest(transportPage(formWithErrors, consignmentReference))),
         validForm => {
```
SteveSugden: You could have extracted consignmentReference into a method (as you started to do) as here, in particular, it only needs to be called if there is an error on the submitted form.
But this is ok as well.

app/views/transport.scala.html
iservport: I believe they should not. Nationality is much shorter than the transport id.

SteveSugden: Fair enough.  They used to be the same size but not on the latest prototype.

PR #204, opened 2019-12-18
CEDS-2027

SteveSugden:
Unified PR for all pages affected by CEDS-2027


app/views/location.scala.html
```diff
@@ -17,27 +17,39 @@
...
+  govukButton: GovukButton,
+  govukInput: GovukInput,
+  exportsInputText: exportsInputText,
+  errorSummary: errorSummary,
+  sectionHeader: sectionHeader,
+  formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF
+)
 
-@(form: Form[Location])(implicit request: Request[_], messages: Messages)
+@(form: Form[Location], consignmentReference: Option[String])(implicit request: JourneyRequest[_], messages: Messages)
```
ed0906: Should consignmentRef always be present at this stage?

app/views/transport.scala.html
```diff
@@ -17,54 +17,121 @@
...
+        govukButton: GovukButton,
+        govukRadios: GovukRadios,
+        govukInput: GovukInput,
+        pageTitle: pageTitle,
+        errorSummary: errorSummary,
+        sectionHeader: sectionHeader,
+        formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF)
 
-@radio(mode: String) = { RadioOption(mode, mode, Transport.messageKey(mode)) }
+@(form: Form[Transport], consignmentReference: Option[String])(implicit request: Request[_], messages: Messages)
```
ed0906: Similar, should consignmentRef always be present here?

SteveSugden: In practice I think it will always be there.

SteveSugden: Do you think it should be changed @ed0906?   Ultimately something has to cope with the possibility that the value might not be there?

ed0906: We could change it in the controller to `consignmentReferences.getOrElse(ReturnToStartException)`

SteveSugden: I forgot this project has that magic exception.    I'll make the changes

ed0906: I think all the UIs have some sort of ReturnToStart exception now

ed0906: You can do it in the view rather than the controller if that's easier. It probably fits more naturally in the controller though

PR #208, opened 2019-12-19
CEDS-2087 Cleaning script references

PatrykRudnicki:
Still `lib/govuk-frontend/govuk/all.js` is looking for `image` that we don't have in assets folder and some fonts. Don't know how to solve those two problems.

Google tag manager is just line that add this script, but we don't have it, so unnecessary thing.

### DIFF
--- a/app/views/components/gds/gds_main_template.scala.html
+++ b/app/views/components/gds/gds_main_template.scala.html
@@ -42,7 +42,6 @@
 @scripts = {
   <script src='@routes.Assets.versioned("lib/govuk-frontend/govuk/all.js")'></script>
   <script src='@routes.Assets.versioned("javascripts/google.tag.manager.min.js")'> </script>
-  <script src="@routes.Assets.versioned("javascripts/autocomplete.js")" type="text/javascript"></script>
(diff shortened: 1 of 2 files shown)


app/views/components/gds/gds_main_template.scala.html
```diff
@@ -41,8 +41,6 @@
 
 @scripts = {
   <script src='@routes.Assets.versioned("lib/govuk-frontend/govuk/all.js")'></script>
-  <script src='@routes.Assets.versioned("javascripts/google.tag.manager.min.js")'> </script>
```
SteveSugden: Is google tags definitely not required then?  What does Emily think?

PatrykRudnicki: I'm not saying is not required. We don't have this script and this causes errors in console.

MaciejRewera: After consideration I agree with Steve. It might be, that this line was added so that we only need to add the google script. But better safe than sorry.

PatrykRudnicki: Yeah, I'll talk to Emily about it

PR #214, opened 2019-12-20
CEDS-2028 - Refactored shut mucr summary page.

iservport:
The reason why this is required is due to the existing frontend assets
library being deprecated by HMRC and GDS and they are no longer
accessible.


app/views/shut_mucr_summary.scala.html
```diff
@@ -14,42 +14,109 @@
...
+        pageTitle: pageTitle,
+        govukSummaryList : GovukSummaryList,
+        formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF)
 
-@(shutMucr: ShutMucr)(implicit request: Request[AnyContent], messages: Messages)
+@(shutMucr: ShutMucr)(implicit request: Request[_], messages: Messages)
 
-@main_template(title = Title("shutMucr.summary.title"), fullWidth = true) {
+@govukLayout(
+    title = Title("disassociate.ucr.summary.title"),
```
bambuchaAdm: is this correct title here ?

app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 HM Revenue & Customs
+ * Copyright 2020 HM Revenue & Customs
```
PatrykRudnicki: Please rebase your branch to the master. As I remember @MaciejRewera updated all the headers

test/unit/views/MovementsViewSpec.scala
```diff
@@ -113,14 +114,20 @@ class MovementsViewSpec extends ViewSpec with Injector {
...
       val secondDataRowElements = pageWithData.selectFirst(".govuk-table__body .govuk-table__row:nth-child(2)")
 
+      val formatter = DateTimeFormatter.ofPattern("dd MMM YYYY 'at' HH:mm")
+
       firstDataRowElements.selectFirst(".ucr").text() mustBe validMucr
       firstDataRowElements.selectFirst(".submission-type").text() mustBe "MUCR"
-      firstDataRowElements.selectFirst(".date-of-request").text() mustBe "31 Oct 2019 at 00:00"
+      firstDataRowElements.selectFirst(".date-of-request").text() mustBe LocalDateTime
+        .of(2019, 10, 31, 0, 0)
+        .format(formatter) // "31 Oct 2019 at 00:00"
```
MaciejRewera: Nit: I see no need for this comment. Formatter is well defined earlier.

iservport: I had an issue with my local machine because of the original hard code String. I had a different locale that broke the test; therefore, my suggestion would be to use this approach to avoid similar problems in the future.

PR #216, opened 2019-12-31
CEDS-2028 Migrate Start page to GDS Design system

PauloPortugal:
**WORK DONE:**

 * **Migrate Start page to GDS Design system**
As part of this work, we kept looking at the prototype, unfortunately 
how the Start page was designed in the prototype it composed of several
`govuk-grid-column-two-thirds` plus some `govuk-grid-column-one-third`.
Since the start page is not a critical page to the service, we have
brought it to the GDS design system, but kept the same layout as before.
Also the Dashed list does not exist as a component in the GDS design
system, so again, we have kept the same approach as before, by simply
using the a normal dash.

* **Import govuk-frontend/govuk assets as the last task**
In order to access all assets and load them correctly, the
`govuk-frontend/govuk` has to be the last imported library in this file.
This actually fixes the problem of not finding the correct fonts for our
Movements FE service, after we migrated to the GDS design system.

### DIFF
--- a/app/views/start_page.scala.html
+++ b/app/views/start_page.scala.html
@@ -14,77 +14,84 @@
  * limitations under the License.
  *@
 
-@import config.AppConfig
 @import views.Title
+@import config.AppConfig
+@import components.gds.gds_main_template
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+@import views.html.components.gds._
 
-@this(main_template: views.html.main_template, appConfig: AppConfig)
+@this(  govukLayout: gds_main_template,
+        govukButton: GovukButton,
+        pageTitle: pageTitle,
+        sectionHeader: sectionHeader,
+        appConfig: AppConfig)
 
 @()(implicit request: Request[_], messages: Messages)
 
-@main_template(title = Title("startPage.title", Some("startPage.title.sectionHeader"))) {
+@reportLink = {<a class="govuk-link" href="@{appConfig.serviceAvailabilityUrl}">@messages("startPage.problemsWithServiceNotice.link")</a>}
+@beforeYouStartLink = {<a class="govuk-link" href=@{appConfig.customsDeclarationsGoodsTakenOutOfEuUrl}>@messages("startPage.beforeYouStart.line.1.link")</a>}
+@govukLayout(title = Title("startPage.title", Some("startPage.title.sectionHeader")), useCustomContentWidth = true) {
 
-  <div class="grid-row">
-    <div class="column-full">
-      @components.heading(
-        sectionHeader = messages("startPage.title.sectionHeader"),
-        title = messages("startPage.title")
-      )
+    <div class="govuk-grid-row">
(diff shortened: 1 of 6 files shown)


app/views/start_page.scala.html
```diff
@@ -14,77 +14,73 @@
...
+
+    <p class="govuk-body-l govuk-!-margin-bottom-8"> @messages("startPage.description") </p>
+
+    <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
+
+    <h2 id="contents" class="govuk-heading-m">@messages("startPage.contents.header")</h2>
+    <ol class="govuk-list">
+        <li> - <a class="govuk-link" href="#before-you-start">@messages("startPage.beforeYouStart.header")</a></li>
+        <li> - <a class="govuk-link" href="#information-you-need">@messages("startPage.informationYouNeed.header")</a></li>
+        <li> - <a class="govuk-link" href="#report-your-arrival-and-departure">@messages("startPage.reportYourArrivalAndDeparture.header")</a></li>
```
PatrykRudnicki: Should we create a class to add this dash automatically to the list?

PauloPortugal:
please test again

PauloPortugal:
test again please

PauloPortugal:
test this please

app/views/start_page.scala.html
bambuchaAdm: We are still able to use this by providing 
```
govuk-list__item--dashed:before {
  content: "-"
}
```
and add correct class to every `li`

bambuchaAdm: Or as later
```
govuk-list--dashed li:before {
  content: "-"
}
```

PauloPortugal: GDS Design System does not have dash list anymore.
But I have made that part of a CSS class. Thank you!

test/unit/views/StartViewSpec.scala
```diff
@@ -76,9 +77,9 @@ class StartViewSpec extends UnitViewSpec {
       view.getElementById("contents").text() mustBe "startPage.contents.header"
 
       view.getElementById("contents-list") must haveChildCount(3)
-      view.getElementById("contents-list").child(0).text() mustBe "startPage.beforeYouStart.header"
-      view.getElementById("contents-list").child(1).text() mustBe "startPage.informationYouNeed.header"
-      view.getElementById("contents-list").child(2).text() mustBe "startPage.reportYourArrivalAndDeparture.header"
+      view.getElementById("contents-list").child(0).text() must include("startPage.beforeYouStart.header")
+      view.getElementById("contents-list").child(1).text() must include("startPage.informationYouNeed.header")
+      view.getElementById("contents-list").child(2).text() must include("startPage.reportYourArrivalAndDeparture.header")
```
PatrykRudnicki: `mustBe` is better than `must include`. Are we able to check the correct message using `mustBe`?

PauloPortugal: Done :)

PR #217, opened 2019-12-31
CEDS-2028 Add govukFlexibleLayout

MaciejRewera:
The default govukLayout provided by play-frontend-govuk does not allow
for customizable page content width. Everything is encapsulated in
govuk-grid-column-two-thirds element.

The new component is a copy of the library's govukLayout but it is
passing the content as is, without adding wrapping elements.
The responsibility for this is moved to the client.

To choose new layout component you have to set useDefaultContentWidth to
false in gds_main_template.

### DIFF
--- a/app/views/components/gds/gds_main_template.scala.html
+++ b/app/views/components/gds/gds_main_template.scala.html
@@ -34,7 +34,7 @@
 @(
   title: Title,
   backButton: Option[BackButton] = None,
-  useDefaultContentWidth: Boolean = true
+  useCustomContentWidth: Boolean = false
 )(contentBlock: Html)(implicit request: Request[_], messages: Messages)
 
 @head = {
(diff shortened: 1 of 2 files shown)


app/views/components/gds/gds_main_template.scala.html
```diff
@@ -19,18 +19,23 @@
...
   siteHeader: siteHeader,
   phaseBanner: phaseBanner
 )
 
-@(  title: Title,
-    backButton: Option[BackButton] = None)(contentBlock: Html)(implicit request: Request[_], messages: Messages)
+@(
+  title: Title,
+  backButton: Option[BackButton] = None,
+  useCustomContentWidth: Boolean = false
```
PauloPortugal: :+1:

PR #218, opened 2019-12-31
CEDS-2028 movements

MaciejRewera:


test/unit/views/MovementsViewSpec.scala
```diff
@@ -20,89 +20,133 @@ import java.time.format.DateTimeFormatter
...
+    val emptyPage = createView()
 
     "contain title" in {
+      emptyPage.getTitle must containMessage("submissions.title")
+    }
+
+    "contain back button" in {
+      emptyPage.getBackButton mustBe defined
+      emptyPage.getBackButton.get must containMessage("site.back.toStartPage")
+      emptyPage.getBackButton.get must haveHref(controllers.routes.ChoiceController.displayChoiceForm())
```
PatrykRudnicki: Instead of checking if the option is defined and after that get, you can use `value` from `OptionValues` to extract data. The error if the option is empty should be legible enough.

non blocking comment

test/unit/views/MovementsViewSpec.scala
```diff
@@ -20,89 +20,133 @@ import java.time.format.DateTimeFormatter
...
+      val allHeaders = emptyPage.getElementsByClass("govuk-table__header")
 
-      page.getElementById("ucr") must containMessage("submissions.ucr")
-      page.getElementById("ucrType") must containMessage("submissions.submissionType")
-      page.getElementById("submissionAction") must containMessage("submissions.submissionAction")
-      page.getElementById("dateOfRequest") must containMessage("submissions.dateOfRequest")
+      allHeaders.get(0) must containMessage("submissions.ucr")
+      allHeaders.get(1) must containMessage("submissions.submissionType")
+      allHeaders.get(2) must containMessage("submissions.dateOfRequest")
+      allHeaders.get(3) must containMessage("submissions.submissionAction")
```
PatrykRudnicki: There is no ID in current elements?

PR #225, opened 2020-01-03
CEDS-2028 notifications

MaciejRewera:


app/assets/stylesheets/customsdecexfrontend-app.scss
```diff
@@ -21,3 +20,5 @@ $govuk-assets-path: '/customs-movements/assets/';
 $govuk-images-path: "/customs-movements/assets/lib/govuk-frontend/govuk/assets/images/";
 $govuk-fonts-path: "/customs-movements/assets/lib/govuk-frontend/govuk/assets/fonts/";
 @import 'lib/govuk-frontend/govuk/all';
+
+@import 'partials/_step-by-step-navigation';
```
PatrykRudnicki: Not sure about it, but probably `@import 'lib/govuk-frontend/govuk/all';` must be the last one. @PauloPortugal please take a look, because you moved it to the end

app/assets/stylesheets/partials/_step-by-step-navigation-overrides.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.exports-notifications__button {
+  cursor: text;
+}
+
+.exports-notifications__header:hover .exports-notifications__circle {
+  color: unset;
+}
+
+.exports-notifications__step:nth-last-child(odd) .exports-notifications__header {
```
bambuchaAdm: you can group those three declarations in `scss` as nested one like
```scss
.exports-notifications__step:nth-last-child(odd) {

  .exports-notifications__header {
    background-color: govuk-colour("light-grey");
  }

  .govuk-form-group {
    background-color: govuk-colour("light-grey");
  }

  .exports-notifications__panel {
    border-bottom: solid 2px govuk-colour("mid-grey");
  }

}
```

app/assets/stylesheets/partials/_step-by-step-navigation-overrides.scss
```diff
@@ -0,0 +1,31 @@
+
+.exports-notifications__header {
+  padding: 15px 0 0 15px;
+  border-top: none;
```
bambuchaAdm: There is syntax to increase DRY
```
.exports-notifications__header {
  padding: 15px 0 0 15px;
  border-top: none;

  &:hover .exports-notifications__circle {
    color: unset;
  }
}
```

app/assets/stylesheets/customsdecexfrontend-app.scss
bambuchaAdm: It have to be just after path variables

MaciejRewera: It has to be like this. The reason being we are using some features from `govuk` so they have to be defined before our code.
By the way it is another thing we have to live with only because we are using design that is not supported for production.

PR #245, opened 2020-02-20
BAU - add Mongock to movements frontend 

SteveSugden:
-  add updated field to existing cache objects

### DIFF
--- a/app/mongock/MongockConfig.scala
+++ b/app/mongock/MongockConfig.scala
@@ -29,7 +29,7 @@ case class MongockConfig(mongoURI: String) {
 
   val runner = new MongockBuilder(client, uri.getDatabase, "mongock.changesets")
     .setLockQuickConfig()
-    .build();
+    .build()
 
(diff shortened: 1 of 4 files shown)


app/mongock/MongockConfig.scala
```diff
@@ -0,0 +1,37 @@
...
+@Singleton
+case class MongockConfig(mongoURI: String) {
+
+  val uri = new MongoClientURI(mongoURI.replaceAllLiterally("sslEnabled", "ssl"))
+
+  val client = new MongoClient(uri)
+
+  val runner = new MongockBuilder(client, uri.getDatabase, "mongock.changesets")
+    .setLockQuickConfig()
+    .build();
```
PatrykRudnicki: Colon at the end is not necessary. It's scala class

SteveSugden: So it is.  I copied that from internal service!

PR #250, opened 2020-02-24
CEDS-2174 - Pre-populate UCR in Movements journeys

iservport:
When the user has searched for a UCR, and then wants to act on it, they should not have to re-enter it.

PLEASE NOTE that I introduced a new ADT, namely `UcrType`, to promote the concept of type MUCR or DUCR to a higher level across the application. I believe it is a good idea if we can incrementally introduce this type to classes where it is needed, instead of other similar types like `AllowedReferences` (which I removed), or like in `AssociateKind` and `DisassociateKind` as untyped Strings.


app/controllers/ChoiceController.scala
```diff
@@ -31,15 +32,24 @@ import views.html.choice_page
...
-      case None          => Ok(choicePage(Choice.form()))
-    }
+  def displayChoiceForm(): Action[AnyContent] = authenticate.async { implicit request =>
+    cacheRepository
+      .findByEori(request.eori)
+      .map {
+        case Some(cache) =>
+          cache.answers
+            .map(answers => Ok(choicePage(Choice.form().fill(Choice(answers.`type`)))))
+            .getOrElse(Ok(choicePage(Choice.form())))
```
MaciejRewera: Actually, I think you should be able to do deep pattern matching here:
`case Some(Cache(_, Some(answers), _, _)) => Ok(choicePage(Choice.form().fill(Choice(answers.`type`))))`

MaciejRewera: Nevermind, you solution will be handy once redirection for no cache is implemented.

app/forms/AssociateUcr.scala
```diff
@@ -50,6 +51,14 @@ object AssociateKind {
 case class AssociateUcr(kind: AssociateKind, ucr: String)
 
 object AssociateUcr {
+
+  def apply(ucrBlock: UcrBlock): AssociateUcr =
+    ucrBlock.ucrType match {
+      case "M" => AssociateUcr(AssociateKind.Mucr, ucrBlock.ucr)
+      case "D" => AssociateUcr(AssociateKind.Ducr, ucrBlock.ucr)
+      case _   => throw new IllegalArgumentException(s"Invalid ucrBlock $ucrBlock")
```
MaciejRewera: Exception message should be rather "Invalid ucrType: $ucrType"

app/models/UcrBlock.scala
```diff
@@ -16,10 +16,17 @@
 
 package models
 
+import forms.UcrType
 import play.api.libs.json.Json
 
-final case class UcrBlock(ucr: String, ucrType: String)
+final case class UcrBlock(val ucr: String, val ucrType: String) {
```
MaciejRewera: `val`s are not needed

app/models/UcrBlock.scala
```diff
@@ -16,10 +16,17 @@
...
+
+  def is(ucrType: UcrType): Boolean = this.ucrType.equals(ucrType.codeValue)
+}
 
 object UcrBlock {
   implicit val format = Json.format[UcrBlock]
+
+  def apply(ucr: String, ucrType: UcrType): UcrBlock = new UcrBlock(ucr, ucrType.codeValue)
```
MaciejRewera: No need to use `new`

conf/app.routes
```diff
@@ -12,6 +12,12 @@ POST        /consignment-query                     controllers.ileQuery.FindCons
 
 GET         /consignment-query/:ucr                controllers.ileQuery.IleQueryController.getConsignmentInformation(ucr: String)
 
+# ILE Query - entry point
+GET         /consignment-query                     controllers.ileQuery.FindConsignmentController.displayQueryForm()
+POST        /consignment-query                     controllers.ileQuery.FindConsignmentController.submitQueryForm()
+
+GET         /consignment-query/:ucr                controllers.ileQuery.IleQueryController.getConsignmentInformation(ucr: String)
+
```
MaciejRewera: Not sure why you have it as your change.

conf/application.conf
```diff
@@ -147,4 +147,10 @@ urls {
   loginContinue = "http://localhost:6796/customs-movements/start"
   customsDeclarationsGoodsTakenOutOfEu = "https://www.gov.uk/guidance/customs-declarations-for-goods-taken-out-of-the-eu"
   serviceAvailability = "https://www.gov.uk/guidance/customs-declaration-service-service-availability-and-issues"
+}
+
+featureSwitches {
+  featureSwitches {
+    queryUcrEnabled = true
+  }
```
MaciejRewera: No need to nest double `featureSwitches`

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -143,6 +144,10 @@ class ChoiceControllerSpec extends ControllerLayerSpec with MockCache {
         redirectLocation(result) mustBe Some(routes.SubmissionsController.displayPage().url)
       }
     }
+
+    "adapt" when {
+      // TODO
+    }
```
MaciejRewera: This does not mean anything. Please remove or implement.

test/unit/controllers/actions/JourneyRefinerSpec.scala
```diff
@@ -43,6 +43,7 @@ class JourneyRefinerSpec extends WordSpec with MustMatchers with MockitoSugar wi
   private val request = AuthenticatedRequest(FakeRequest(), user)
   private val answers = AssociateUcrAnswers()
   private val cache = Cache("eori", answers)
+  private val redirectResult = Results.Redirect(controllers.routes.ChoiceController.displayChoiceForm())
```
MaciejRewera: Nit: In our projects we usually don't write `Result` before `Redirect`

test/unit/controllers/actions/JourneyRefinerSpec.scala
```diff
@@ -52,14 +53,14 @@ class JourneyRefinerSpec extends WordSpec with MustMatchers with MockitoSugar wi
   }
 
   "refine" should {
-    "permit request" when {
-      "answers found" in {
+    "add cache to the request" when {
+      "cache and answers are both not empty" in {
```
MaciejRewera: I'd re-phrase this line. If answers are non-empty, then cache per-se is non-empty as well.
Maybe something like "cache contains non-empty answers" ?

test/unit/controllers/actions/JourneyRefinerSpec.scala
```diff
@@ -70,10 +71,16 @@ class JourneyRefinerSpec extends WordSpec with MustMatchers with MockitoSugar wi
     }
 
     "block request" when {
-      "answers not found" in {
+
+      "cache is empty" in {
```
MaciejRewera: In this case cache does not exist. This is different from being empty.

MaciejRewera: Actually, this test case is not finished.

test/unit/views/ViewSpec.scala
```diff
@@ -27,13 +27,17 @@ import play.twirl.api.Html
...
 
 class ViewSpec extends WordSpec with MustMatchers with ViewMatchers with MessagesStub with CSRFSupport {
 
   implicit protected def htmlBodyOf(html: Html): Document = Jsoup.parse(html.toString())
 
   protected def journeyRequest(answers: Answers) =
-    JourneyRequest(answers, AuthenticatedRequest(FakeRequest().withCSRFToken, SignedInUser("eori", Enrolments(Set.empty))))
+    JourneyRequest(
+      Cache(validEori, Some(answers), None),
+      AuthenticatedRequest(FakeRequest().withCSRFToken, SignedInUser("eori", Enrolments(Set.empty)))
```
MaciejRewera: Please use the same eori as it is defined for `AuthenticatedRequest` element

test/unit/controllers/actions/JourneyRefinerSpec.scala
iservport: The line does not compile with error `not found: value Redirect` when `Result.Redirect` is  not used.

app/controllers/ChoiceController.scala
SteveSugden: There was no need to change this method.  As indicated in your TODO, I've done/am doing this as part of 2175

app/forms/DisassociateUcr.scala
```diff
@@ -55,6 +55,13 @@ case class DisassociateUcr(kind: DisassociateKind, ducr: Option[String], mucr: O
 object DisassociateUcr {
   import uk.gov.voa.play.form.ConditionalMappings._
 
+  def apply(ucrBlock: UcrBlock): DisassociateUcr =
+    ucrBlock.ucrType match {
+      case "M" => DisassociateUcr(DisassociateKind.Mucr, None, Some(ucrBlock.ucr))
+      case "D" => DisassociateUcr(DisassociateKind.Ducr, None, Some(ucrBlock.ucr))
+      case _   => throw new IllegalArgumentException(s"Invalid ucrBlock $ucrBlock")
```
MaciejRewera: `s"Invalid ucrType: ${ucrBlock.ucrType}"`

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -18,9 +18,10 @@ package unit.controllers
 
 import controllers.consolidations.{routes => consolidationRoutes}
 import controllers.{routes, ChoiceController}
-import forms.Choice
+import forms.{AssociateKind, AssociateUcr, Choice, ConsignmentReferences, DisassociateKind, DisassociateUcr, ShutMucr, UcrType}
```
MaciejRewera: Nit: in such cases it is safer to use `forms._` because sometimes code formatters are complaining about the order.

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -89,49 +91,114 @@ class ChoiceControllerSpec extends ControllerLayerSpec with MockCache {
...
-        val result = controller.submitChoice()(postRequest(arrivalForm))
+        "create cache if empty" in {
+          givenTheCacheIsEmpty()
+          val result = controller.submitChoice()(postRequest(arrivalForm))
 
-        status(result) mustBe SEE_OTHER
-        redirectLocation(result) mustBe Some(routes.ConsignmentReferencesController.displayPage().url)
+          status(result) mustBe SEE_OTHER
+          redirectLocation(result) mustBe Some(routes.ConsignmentReferencesController.displayPage().url)
+          theCacheUpserted.answers mustBe Some(ArrivalAnswers(None, None, None))
```
MaciejRewera: Nit: Aren't `None`s the default values for empty `ArrivalAnswers`?

iservport: Yes, I made them explicit to reinforce the use of None in each case.

PR #251, opened 2020-02-25
CEDS-2175 - Make searching for a consignment the first step

SteveSugden:
This PR provides the following
- 'Start' button on /start directs to either 'Find a consignment' or 'Choice' page depending on feature switch
- 'Choice' page with feature disabled - behaves as before; pre-populates if answer present; 'back' link to 'Start' page.
- 'Choice' page with feature enabled - redirects to 'Find a consignment' if queried ucr not present otherwise populates with answer if present; 'back' links to previous search.


app/views/components/links/ChoicePageBackLink.scala
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package views.components.links
+import models.UcrBlock
+import play.api.mvc.Call
+
+object ChoicePageBackLink {
+  def call(ileQueryEnabled: Boolean, queryUcr: Option[UcrBlock]): Call =
```
bambuchaAdm: This flag could be injected and this object could be moved as injeted class to view - this would does not require to pass flags and appConfig to view

PR #253, opened 2020-02-26
CEDS-2057 - Rename the Export services

iservport:
A consistent naming convention must be displayed across all three CDS Exports services to provide users for a simplified experience and understanding.

The CDS Movements Exports service was modified to:

1) the service name changed to CDS Exports in the banners at the top of each page,
2) the service name changed to CDS Exports on the start page.

### DIFF
--- a/conf/messages
+++ b/conf/messages
@@ -156,7 +156,7 @@ movement.departure = Depart goods
 
 startPage.title.sectionHeader = Guidance
 startPage.title = Tell HMRC when goods you’re exporting arrive at or leave the UK port
-startPage.description = Use the CDS Esports Service to tell HMRC when the goods you’re exporting out of the EU have arrived at or left the UK port.
+startPage.description = Use the CDS Exports Service to tell HMRC when the goods you’re exporting out of the EU have arrived at or left the UK port.
 startPage.contents.header = Contents
 startPage.beforeYouStart.header = Before you start
 startPage.beforeYouStart.line.1 = The goods must have already been {0}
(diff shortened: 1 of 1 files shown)


conf/messages
```diff
@@ -165,7 +156,7 @@ movement.departure = Depart goods
 
 startPage.title.sectionHeader = Guidance
 startPage.title = Tell HMRC when goods you’re exporting arrive at or leave the UK port
-startPage.description = Use the Customs Declaration Service to tell HMRC when the goods you’re exporting out of the EU have arrived at or left the UK port.
+startPage.description = Use the CDS Esports Service to tell HMRC when the goods you’re exporting out of the EU have arrived at or left the UK port.
```
PatrykRudnicki: Typo in `Exports`. You have `Esports`

SteveSugden: typo - exports

conf/messages
```diff
@@ -34,15 +34,6 @@ error.ducr.format = DUCR number is in incorrect format
 error.mucr.empty = MUCR number cannot be empty
 error.mucr.format = MUCR number is in incorrect format
 
-index.title =Export Declarations
```
SteveSugden: These are indeed un-used

SteveSugden:
Have you checked acceptance tests BTW?

conf/messages
iservport: I could not find any page that still uses message keys like index.*.

SteveSugden: Indeed, you are correct.  I double checked to save anyone else doing so.

PR #256, opened 2020-02-27
CEDS-2057 - Rename the Export services

iservport:
Fix banners.

A consistent naming convention must be displayed across all three CDS Exports services to provide users for a simplified experience and understanding.

The CDS Movements Exports service was modified to:

1) the service name changed to CDS Exports in the banners at the top of each page,
2) the service name changed to CDS Exports on the start page.

### DIFF
--- a/test/unit/views/components/SiteHeaderSpec.scala
+++ b/test/unit/views/components/SiteHeaderSpec.scala
@@ -1,0 +1,24 @@
+package views.components
+
+import base.Injector
+import play.api.test.FakeRequest
+import play.twirl.api.Html
+import views.ViewSpec
+import views.html.components.gds.siteHeader
+
+class SiteHeaderSpec extends ViewSpec with Injector {
+
+  private val page = instanceOf[siteHeader]
+  private implicit val request = FakeRequest()
+
+  private def createHeader(): Html = page()(request, messages)
+
+  "Site header" should {
+    val siteHeader = createHeader()
+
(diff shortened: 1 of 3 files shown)


test/unit/views/components/SiteHeaderSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+  private val page = instanceOf[siteHeader]
+  private implicit val request = FakeRequest()
+
+  private def createHeader(): Html = page()(request, messages)
+
+  "Site header" should {
+    val siteHeader = createHeader()
+
+    "display banner with the service name" in {
+      siteHeader.getElementsByClass("govuk-header__link govuk-header__link--service-name").first() must containMessage("service.name")
```
bambuchaAdm: for 99% you could use only modifier part on class -

test/unit/views/components/SiteHeaderSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+import base.Injector
+import play.api.test.FakeRequest
+import play.twirl.api.Html
+import views.ViewSpec
+import views.html.components.gds.siteHeader
+
+class SiteHeaderSpec extends ViewSpec with Injector {
+
+  private val page = instanceOf[siteHeader]
+  private implicit val request = FakeRequest()
```
PatrykRudnicki: Request doesn't need to be implicit if you're passing it explicite anyway

PR #259, opened 2020-03-02
CEDS-2183 Remove View Requests choice

PatrykRudnicki:


conf/application.conf
```diff
@@ -148,7 +148,3 @@ urls {
   customsDeclarationsGoodsTakenOutOfEu = "https://www.gov.uk/guidance/customs-declarations-for-goods-taken-out-of-the-eu"
   serviceAvailability = "https://www.gov.uk/guidance/customs-declaration-service-service-availability-and-issues"
 }
-
-featureSwitches {
```
SteveSugden: 👍

app/views/choice_page.scala.html
```diff
@@ -35,6 +35,43 @@
 
 @(form: Form[Choice], queryUcr: Option[UcrBlock] = None)(implicit request: Request[_], messages: Messages)
 
+@choices = @{
```
SteveSugden: I would move the creation of this sequence into the ChoicePageConfig helper.   Currently, you don't seem to have any tests for the contents of the Seq other than switching based on the real app config.  If the method was in the helper you could test used the mocked appConfig approach.

PatrykRudnicki: I agree, but unfortunately every option is `RadioItem` which requires `Text` , `messages` and `form`. In general mowing this to the `ChoicePageConfig` can make implementation slightly more complicated.

SteveSugden: True.  How about this - using the ChoicePageConfig to provide a filter that can remove choice items depending on feature flag.  Then the page can construct the RadioItems and the PageConfig can refine.   At the moment, the only thing you can test on ChoicePageConfig is that pageConfig.isQueryEnabled == true when it's defined that way in application.config.

PatrykRudnicki: Hmm, so list of the all choice in view and after that method which will filter the incorrect options?

PatrykRudnicki: Added more tests around it

PR #260, opened 2020-03-03
CEDS-2184 - Prevent changes to UCR when Managing Consignment

SteveSugden:
This ticket prevents the user changing the queried ucr on all journeys except associate.
Also prevents access to the "change ucr" page when ileQuery flag enabled.


test/it/ArrivalSpec.scala
```diff
@@ -32,21 +33,23 @@ class ArrivalSpec extends IntegrationSpec {
 
   "Consignment References Page" when {
     "GET" should {
-      "return 200" in {
+      "throw feature disabled" in {
```
PatrykRudnicki: I think for this test case we need to manually build `AppConfig` to always have a `ileQuery` enabled, because when someone will disable this locally and run tests it will fail, because there is no exception.

But I'm not sure if it's a mistake, because I can't imagine a situation when we will need to disable this locally, but it's somehow possible

test/it/ArrivalSpec.scala
```diff
@@ -32,21 +33,23 @@ class ArrivalSpec extends IntegrationSpec {
...
 
         // When
         val response = get(controllers.routes.ConsignmentReferencesController.displayPage())
 
-        // TThen
-        status(response) mustBe OK
+        // Then
+        intercept[RuntimeException] {
+          status(response)
+        } mustBe FeatureDisabledException
```
PatrykRudnicki: Here I think better is use:
```
intercept[FeatureDisabledException] {
    await(response)
}
```

`await` is from play framework so you don't need to extends any additional traits etc. Additionally the reason of using the `status` is to check the status, here we need to just wait for the result. Of course actual implementation works as expected.

test/it/ArrivalSpec.scala
```diff
@@ -56,9 +59,9 @@ class ArrivalSpec extends IntegrationSpec {
           post(controllers.routes.ConsignmentReferencesController.saveConsignmentReferences(), "reference" -> "M", "mucrValue" -> "GB/123-12345")
 
         // Then
-        status(response) mustBe SEE_OTHER
-        redirectLocation(response) mustBe Some(controllers.routes.MovementDetailsController.displayPage().url)
-        theAnswersFor("eori") mustBe Some(ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences("M", "GB/123-12345"))))
+        intercept[RuntimeException] {
+          status(response)
+        } mustBe FeatureDisabledException
```
PatrykRudnicki: The same here and in every place where you added expecting the exception

PatrykRudnicki:
Please remember about squashing commits before merge and you have some build failure.

app/controllers/actions/NonIleQueryAction.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import config.AppConfig
+import controllers.exception.FeatureDisabledException
+import javax.inject.Inject
+import models.requests.AuthenticatedRequest
+import play.api.mvc._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class NonIleQueryAction @Inject()(appConfig: AppConfig)(implicit val exc: ExecutionContext)
```
MaciejRewera: Bear in mind that this is another class that we'll need to remove once the feature flag is not needed.
An alternative could be to parameterize existing `IleQueryAction`

test/it/ArrivalSpec.scala
PatrykRudnicki: And additionally it's good to know that the result of the intercept is an exception, so if you need to validate a message from exception you can put the result to the val and check the message.

This is the reason that using `mustBe` here is working correctly

app/views/components/config/DissociateSummaryConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package views.components.config
+
+import config.AppConfig
+import javax.inject.Inject
+import play.api.mvc.Call
+
+class DissociateSummaryConfig @Inject()(appConfig: AppConfig) {
```
MaciejRewera: How about instead of creating several ...Config classes, create something like a Navigator in Declarations service?

SteveSugden: I thought the idea was to limit the areas of concern - i.e. one class to determine the behaviour of one view.   The code will be easy to remove.  Delete the feature flag in AppConfig and fix all the compilation errors.

app/controllers/actions/NonIleQueryAction.scala
SteveSugden: Could do.   But if the parameter is a boolean it might get confusing as to which way it should be.  Again, classes are easy to delete when they're no longer needed.

app/controllers/actions/NonIleQueryAction.scala
```diff
@@ -0,0 +1,34 @@
...
+    extends ActionFunction[AuthenticatedRequest, AuthenticatedRequest] {
+
+  override def invokeBlock[A](request: AuthenticatedRequest[A], block: AuthenticatedRequest[A] => Future[Result]): Future[Result] =
+    if (appConfig.ileQueryEnabled) throw FeatureDisabledException else block(request)
```
MaciejRewera: BTW, in my opinion `FeatureDisabledException` is not a perfect name for it, as it gets ambiguous. What a "Feature" means here? Because by the way it is used it is clear that it is not the same as "ILE Query feature".

test/it/ArrivalSpec.scala
SteveSugden: IleQuery is enabled by default in `IntegrationSpec` base class.

PatrykRudnicki: In that case it's enough :)

app/controllers/actions/NonIleQueryAction.scala
PatrykRudnicki: I think that this logic must stay after the feature flag is removed. I mean if we don't have feature flag user shouldn't be able to access this page.

app/controllers/actions/NonIleQueryAction.scala
SteveSugden: I see what you mean.  It's really a "FeatureENABLEDException" in this case.

Maybe it should be more generically named - something like "InvalidAccess".   Suggestions?

test/it/IleQueryDisabled.scala
```diff
@@ -0,0 +1,6 @@
+import play.api.Configuration
+trait IleQueryDisabled extends IntegrationSpec {
```
MaciejRewera: IMHO this is a code smell. I understand that this trait is simply not to repeat yourself.
But in every Spec both `IleQueryDisabled` and `IntegrationSpec` are mixed-in.
I'd rather override the method in every Spec it is needed.

app/controllers/actions/NonIleQueryAction.scala
SteveSugden: Once the feature flag is removed, we can delete the page that we are trying to protect (just like we are currently doing in the internal version of the service).

PatrykRudnicki: Yeah, you're right

test/it/IleQueryDisabled.scala
SteveSugden: I started doing that and after the 3rd or 4th time I thought, this would look better with a "with.... " trait.    

There's no pleasing some people.

test/it/ArrivalSpecWithIleQueryDisabled.scala
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+import java.time.temporal.ChronoUnit
+import java.time.{LocalDate, LocalDateTime, LocalTime, ZoneOffset}
+
+import forms.ConsignmentReferences
+import models.cache.ArrivalAnswers
+import play.api.test.Helpers._
+
+class ArrivalSpecWithIleQueryDisabled extends IntegrationSpec with IleQueryDisabled {
```
MaciejRewera: I am not convinced about separate Spec just to test different configuration. I am mostly concerned about approach like this, having in mind that we will have to clear everything after the feature flag is not used.
I'd probably put both scenarios into single test suite, but changed the way `ileQueryFeatureConfiguration` value is obtained, to make it parameterized.

test/it/DissociateUcrSpec.scala
```diff
@@ -46,11 +48,9 @@ class DissociateUcrSpec extends IntegrationSpec {
         val response = post(controllers.consolidations.routes.DisassociateUcrController.submit(), "kind" -> "mucr", "mucr" -> "GB/321-54321")
 
         // Then
-        status(response) mustBe SEE_OTHER
-        redirectLocation(response) mustBe Some(controllers.consolidations.routes.DisassociateUcrSummaryController.displayPage().url)
-        theAnswersFor("eori") mustBe Some(
-          DisassociateUcrAnswers(ucr = Some(DisassociateUcr(kind = DisassociateKind.Mucr, mucr = Some("GB/321-54321"), ducr = None)))
-        )
+        intercept[RuntimeException] {
+          status(response)
```
MaciejRewera: Where is an assertion? Is it on purpose?

test/it/ShutMucrSpec.scala
```diff
@@ -46,9 +48,9 @@ class ShutMucrSpec extends IntegrationSpec {
         val response = post(controllers.consolidations.routes.ShutMucrController.submitForm(), "mucr" -> "GB/123-12345")
 
         // Then
-        status(response) mustBe SEE_OTHER
-        redirectLocation(response) mustBe Some(controllers.consolidations.routes.ShutMucrSummaryController.displayPage().url)
-        theAnswersFor("eori") mustBe Some(ShutMucrAnswers(shutMucr = Some(ShutMucr("GB/123-12345"))))
+        intercept[RuntimeException] {
+          status(response)
```
MaciejRewera: Similar here

test/it/DepartureSpec.scala
```diff
@@ -56,9 +58,9 @@ class DepartureSpec extends IntegrationSpec {
           post(controllers.routes.ConsignmentReferencesController.saveConsignmentReferences(), "reference" -> "M", "mucrValue" -> "GB/123-12345")
 
         // Then
-        status(response) mustBe SEE_OTHER
-        redirectLocation(response) mustBe Some(controllers.routes.MovementDetailsController.displayPage().url)
-        theAnswersFor("eori") mustBe Some(DepartureAnswers(consignmentReferences = Some(ConsignmentReferences("M", "GB/123-12345"))))
+        intercept[RuntimeException] {
+          status(response)
```
MaciejRewera: Where is an assertion?

test/it/DissociateUcrSpec.scala
SteveSugden: No, it's a mistake.  Addressing with Patryk's comment

test/unit/controllers/consolidations/DisassociateUcrControllerSpec.scala
```diff
@@ -37,20 +40,30 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
   override protected def beforeEach(): Unit = {
     super.beforeEach()
-    when(page.apply(any())(any(), any())).thenReturn(HtmlFormat.empty)
+    when(page.apply(any())(any(), any())) thenReturn HtmlFormat.empty
+    when(config.ileQueryEnabled) thenReturn false
   }
 
   override protected def afterEach(): Unit = {
     reset(page)
+    reset(config)
```
MaciejRewera: This can be expressed in a single line: `reset(page, config)`

test/unit/views/ArrivalDetailsViewSpec.scala
```diff
@@ -62,7 +66,12 @@ class ArrivalDetailsViewSpec extends ViewSpec with Injector {
         val backButton = emptyView.getBackButton
 
         backButton mustBe defined
-        backButton.get must haveHref(controllers.routes.ConsignmentReferencesController.displayPage())
+        backButton.get must haveHref(
+          if (realAppConfig.ileQueryEnabled)
+            controllers.routes.ChoiceController.displayChoiceForm()
+          else
+            controllers.routes.ConsignmentReferencesController.displayPage()
+        )
```
MaciejRewera: IMO all cases should be tested every time on unit level. So instead of using real app config, I'd advise to use a mock and have 2 test cases depending on whether the feature flag is enabled or not.

test/unit/views/DepartureDetailsViewSpec.scala
```diff
@@ -62,7 +65,12 @@ class DepartureDetailsViewSpec extends ViewSpec with Injector {
         val backButton = emptyView.getBackButton
 
         backButton mustBe defined
-        backButton.get must haveHref(controllers.routes.ConsignmentReferencesController.displayPage())
+        backButton.get must haveHref(
+          if (realAppConfig.ileQueryEnabled)
+            controllers.routes.ChoiceController.displayChoiceForm()
+          else
+            controllers.routes.ConsignmentReferencesController.displayPage()
+        )
```
MaciejRewera: Same here. IMO there should be 2 test cases.

test/unit/views/components/config/IleQueryFeatureConfigSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package views.components.config
+import base.UnitSpec
+import config.AppConfig
+import org.mockito.Mockito.when
+import org.scalatest.BeforeAndAfterEach
+
+trait IleQueryFeatureConfigSpec extends UnitSpec with BeforeAndAfterEach {
```
MaciejRewera: I think that extending `UnitSpec` here is an overkill. `MockitoSugar` should be sufficient.

test/unit/views/ArrivalDetailsViewSpec.scala
SteveSugden: I was wondering about that when I did that (not in this PR).
Really, the href doesn't need to be tested here because it's tested in the config helper class that generates the url.  It should be enough to test that there _is_ a back link and that's all.

test/unit/views/components/config/IleQueryFeatureConfigSpec.scala
SteveSugden: I find that's the problem with mixing in all these traits - working out which one you really need.  So I go with the one that works!   But I'll see if you're right.

SteveSugden:
arrival_details view is now using the `MovementDetailsConfig` to generate the back button

app/views/components/config/DissociateSummaryConfig.scala
MaciejRewera: Didn't think about it in this way. I agree that with this in mind it looks more sensible.

test/it/IleQueryDisabled.scala
MaciejRewera: I see. But then it makes no point in mixing both in concrete Specs.

test/unit/views/ArrivalDetailsViewSpec.scala
MaciejRewera: In that case I'd do exactly as you said: "...to test that there is a back link and that's all", but without checking the urls. If it was like that, then I'd expect that the url itself is tested somewhere else. But ATM the intention is not clear to me as a reader.

test/it/DepartureSpec.scala
```diff
@@ -56,9 +59,9 @@ class DepartureSpec extends IntegrationSpec {
           post(controllers.routes.ConsignmentReferencesController.saveConsignmentReferences(), "reference" -> "M", "mucrValue" -> "GB/123-12345")
 
         // Then
-        status(response) mustBe SEE_OTHER
-        redirectLocation(response) mustBe Some(controllers.routes.MovementDetailsController.displayPage().url)
-        theAnswersFor("eori") mustBe Some(DepartureAnswers(consignmentReferences = Some(ConsignmentReferences("M", "GB/123-12345"))))
+        intercept[InvalidFeatureStateException] {
+          await(response)
+        }
```
MaciejRewera: 👍

test/it/ShutMucrSpec.scala
```diff
@@ -23,7 +24,7 @@ class ShutMucrSpec extends IntegrationSpec {
 
   "Enter MUCR Page" when {
     "GET" should {
-      "return 200" in {
+      "throw feature disabled" in {
```
MaciejRewera: Nit: now it throws Invalid Feature State ATM. But this is not that important here.

app/controllers/actions/NonIleQueryAction.scala
```diff
@@ -0,0 +1,34 @@
...
+import models.requests.AuthenticatedRequest
+import play.api.mvc._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class NonIleQueryAction @Inject()(appConfig: AppConfig)(implicit val exc: ExecutionContext)
+    extends ActionFunction[AuthenticatedRequest, AuthenticatedRequest] {
+
+  override def invokeBlock[A](request: AuthenticatedRequest[A], block: AuthenticatedRequest[A] => Future[Result]): Future[Result] =
+    if (appConfig.ileQueryEnabled) throw InvalidFeatureStateException else block(request)
```
bambuchaAdm: Why exactly this exception ?

test/unit/controllers/consolidations/DisassociateUcrControllerSpec.scala
```diff
@@ -37,20 +40,29 @@ import scala.concurrent.ExecutionContext.Implicits.global
 class DisassociateUcrControllerSpec extends ControllerLayerSpec with MockCache with ScalaFutures with OptionValues {
 
   private val page = mock[disassociate_ucr]
+  private val config = mock[AppConfig]
```
bambuchaAdm: why we would like to mock this ? Can we provide any fixed behaviour stub for action in tests ?

test/unit/views/components/config/IleQueryFeatureConfigSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  val ileQueryEnabled = mock[AppConfig]
+  val ileQueryDisabled = mock[AppConfig]
+
+  override def beforeEach() {
+    super.beforeEach()
```
bambuchaAdm: please add resets on `afterEach`

PR #263, opened 2020-03-06
CEDS-2131 Unit Tests refactoring

MaciejRewera:
This PR fixes some unit tests and changes ConsignmentReference to be val in Controllers used for Arrival/Departure journey.

### DIFF
--- a/app/controllers/MovementDetailsController.scala
+++ b/app/controllers/MovementDetailsController.scala
@@ -52,7 +52,10 @@ class MovementDetailsController @Inject()(
   }
 
   private def arrivalPage(arrivalAnswers: ArrivalAnswers)(implicit request: JourneyRequest[AnyContent]): Html =
-    arrivalDetailsPage(arrivalAnswers.arrivalDetails.fold(details.arrivalForm())(details.arrivalForm().fill(_)), arrivalAnswers.consignmentReferences)
+    arrivalDetailsPage(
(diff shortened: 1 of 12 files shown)


app/controllers/MovementDetailsController.scala
```diff
@@ -72,24 +75,26 @@ class MovementDetailsController @Inject()(
...
+        validForm =>
+          cache.upsert(request.cache.update(arrivalAnswers.copy(arrivalDetails = Some(validForm)))).map { _ =>
+            Right(controllers.routes.LocationController.displayPage())
+        }
+      )
   }
 
   private def handleSavingDeparture(departureAnswers: DepartureAnswers)(implicit request: JourneyRequest[AnyContent]): Future[Either[Html, Call]] = {
-    def consignmentReference = departureAnswers.consignmentReferences.map(_.referenceValue).getOrElse(throw ReturnToStartException)
+    val consignmentReferenceValue = departureAnswers.consignmentReferences.map(_.referenceValue).getOrElse(throw ReturnToStartException)
```
bambuchaAdm: why rename to `value` ?

MaciejRewera: `ConsignmentReference` is a case class, but the value under this field is of type `String`. I was confused when started working in this area, so decided that it is better to explicitly state that this is not an instance of `ConsignmentReference`.
I thought about calling it just `ucr`, because essentially it is a UCR. But adding "Value" as a suffix introduces less changes.

PR #264, opened 2020-03-06
CEDS-2201 - Prevent changes to UCR when Associating

SteveSugden:
This PR prevents the user changing the queried Ucr when on the Associate journey.
Includes the new ability to associate queried Mucr with another Mucr or vice verse.
Includes new summary page for Associate journey in line with latest prototype.

### DIFF
--- a/app/controllers/consolidations/AssociateUcrSummaryController.scala
+++ b/app/controllers/consolidations/AssociateUcrSummaryController.scala
@@ -16,8 +16,10 @@
 
 package controllers.consolidations
 
+import config.AppConfig
 import controllers.actions.{AuthAction, JourneyRefiner}
 import controllers.storage.FlashKeys
+import forms.{AssociateKind, ManageMucrChoice}
 import javax.inject.{Inject, Singleton}
 import models.ReturnToStartException
 import models.cache.{AssociateUcrAnswers, JourneyType}
@@ -25,7 +27,7 @@ import play.api.i18n.I18nSupport
 import play.api.mvc._
 import services.SubmissionService
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
-import views.html.associateucr.associate_ucr_summary
+import views.html.associateucr.{associate_ucr_summary, associate_ucr_summary_no_change}
 
 import scala.concurrent.ExecutionContext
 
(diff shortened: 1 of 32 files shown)


app/views/components/config/MucrOptionsConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+class MucrOptionsConfig @Inject()(appConfig: AppConfig) extends BaseConfig(appConfig) {
+
+  def backUrl(manageMucrChoice: Option[ManageMucrChoice] = None): Call =
+    if (appConfig.ileQueryEnabled)
+      if (manageMucrChoice.isDefined)
+        controllers.consolidations.routes.ManageMucrController.displayPage()
+      else
+        controllers.routes.ChoiceController.displayChoiceForm()
+    else
+      controllers.routes.ChoiceController.displayChoiceForm()
```
PatrykRudnicki:
```
if(appConfig.ileQueryEnabled && manageMucrChoice.isDefined) 
controllers.consolidations.routes.ManageMucrController.displayPage()
else
controllers.routes.ChoiceController.displayChoiceForm()
```

Small propose to avoid nested ifs

SteveSugden: There are three outcomes here, not two.  Maybe some java-style brackets would help

PatrykRudnicki: Yeah, but two of them are the same. Those two elses contains the same redirection

SteveSugden: True- didn't spot that.  I was more focused on keeping the "ileQuery feature" behaviour clearly separated from the "non ileQuey feature" (so it's simpler to remove).  But on reflection, you are right.

app/controllers/consolidations/AssociateUcrSummaryController.scala
```diff
@@ -35,15 +37,26 @@ class AssociateUcrSummaryController @Inject()(
...
 
   def displayPage(): Action[AnyContent] = (authenticate andThen journeyType(JourneyType.ASSOCIATE_UCR)) { implicit request =>
     val answers = request.answersAs[AssociateUcrAnswers]
     val mucrOptions = answers.mucrOptions.getOrElse(throw ReturnToStartException)
-    val ucr = answers.associateUcr.getOrElse(throw ReturnToStartException)
-    Ok(associateUcrSummaryPage(ucr, mucrOptions.mucr))
+    val associateUcr = answers.associateUcr.getOrElse(throw ReturnToStartException)
+
+    if (appConfig.ileQueryEnabled) {
+      val associateAnotherMucr = answers.manageMucrChoice.exists(_.choice == ManageMucrChoice.AssociateAnotherMucr)
```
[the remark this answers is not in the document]
SteveSugden: I'm moving that test to a method in the answers class - called `isAssociateAnotherMucr`

PR #266, opened 2020-03-11
CEDS-2210 - Allow Loaders to use the current dateTime when arriving or departing consignments

SteveSugden:


SteveSugden:
I always squash Patryk.   But your commit is a mystery.   It was made days ago and I only started this branch yesterday afternoon (after doing a fresh "pull" from master).   Strange.

PR #274, opened 2020-05-28
CEDS-2434 - content changes

SteveSugden:


app/config/AppConfig.scala
```diff
@@ -55,6 +55,7 @@ class AppConfig @Inject()(
   lazy val loginUrl = loadConfig("urls.login")
   lazy val loginContinueUrl = loadConfig("urls.loginContinue")
   lazy val customsDeclarationsGoodsTakenOutOfEuUrl = loadConfig("urls.customsDeclarationsGoodsTakenOutOfEu")
+  lazy val tradeTariffUrl = loadConfig("urls.tradeTariff")
```
PatrykRudnicki, line 58: I think that you're not using this.

You have this created in `MucrOptionsConfig.scala`

SteveSugden: MucrOptionsConfig uses AppConfig

PatrykRudnicki: Ah, yeah, I though that we're retrieving this from config just in MucrOptionsConfig

PR #275, opened 2020-06-03
CEDS-2464 Sign out link

MaciejRewera:


app/assets/stylesheets/customsdecexfrontend-app.scss
```diff
@@ -19,7 +19,10 @@
 $govuk-assets-path: '/customs-movements/assets/';
 $govuk-images-path: "/customs-movements/assets/lib/govuk-frontend/govuk/assets/images/";
 $govuk-fonts-path: "/customs-movements/assets/lib/govuk-frontend/govuk/assets/fonts/";
+$hmrc-assets-path: "/self-employment-support/assets/lib/hmrc-frontend/hmrc/assets/";
```
SteveSugden, line 22: Self employment?

app/models/SignOutReason.scala
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+
+import play.api.mvc.QueryStringBindable
+
+sealed abstract class SignOutReason(val name: String)
```
SteveSugden, line 21: I am always amazed by what a pain "enums" are in Scala and the hoops you have to jump thought to serialise them.

What's wrong with a nice String?  (Not suggesting you change it - just having a rant)

app/views/components/gds/gds_main_template.scala.html
```diff
@@ -38,7 +38,8 @@
 @(
   title: Title,
   backButton: Option[BackButton] = None,
-  useCustomContentWidth: Boolean = false
+  useCustomContentWidth: Boolean = false,
+  showSignOutLink: Boolean = true
```
SteveSugden, line 42: Could we infer this variable (showSignOutLink) based on whether the user is logged in or not (like the useTimeoutDialog method below)?

app/views/components/gds/siteHeader.scala.html
```diff
@@ -14,17 +14,31 @@
...
 
-@this(govukHeader: GovukHeader)
+@this(hmrcHeader: HmrcHeader)
 
-@()(implicit request: Request[_], messages: Messages)
+@(showSignOut: Boolean = true)(implicit messages: Messages)
 
-@govukHeader(Header(
-    homepageUrl = None,
+@languageSelected = @{ if(messages.lang.code == "cy")
```
SteveSugden, line 25: Are we supporting Welsh now?

test/unit/views/ArrivalSummaryViewSpec.scala
```diff
@@ -74,75 +74,91 @@ class ArrivalSummaryViewSpec extends ViewSpec with MockitoSugar with BeforeAndAf
...
+      val answer_consignment_reference_link_index = answer_consignment_reference + 1
+
       val view = page(answers)
       view.getElementsByClass("govuk-heading-m").get(section_consignment_details) must containMessage("summary.consignmentDetails")
 
       view.getElementsByClass("govuk-summary-list__key").get(answer_consignment_type) must containMessage("summary.referenceType")
       view.getElementsByClass("govuk-summary-list__value").get(answer_consignment_type) must containMessage("consignmentReferences.reference.ducr")
 
-      val changeType = view.getElementsByClass("govuk-link").get(answer_consignment_type)
+      val changeType = view.getElementsByClass("govuk-link").get(answer_consignment_type_link_index)
```
SteveSugden, line 97: Ok dear.  I presume all of this "link" tests have been thrown out by the introduction of a new sign out link at the top of the page?

app/views/components/gds/siteHeader.scala.html
MaciejRewera: As @PauloPortugal told me, we are going to implement it in the next ~3 months. I'm not going to be an advocate to keep it now, but I'd like to know what everyone thinks about it.

test/unit/views/ArrivalSummaryViewSpec.scala
MaciejRewera: Unfortunately yes :/

app/views/components/gds/siteHeader.scala.html
SteveSugden: What happens if you select Welsh?  Does it default to English?

app/models/SignOutReason.scala
MaciejRewera: Strings are too generic in my opinion. Prefer to have types if only possible. Using type here limits allowed values and also allows for 'cleaner' pattern matching in controller.
Besides, it requires more code not because of being "enum" but because this type is used as URL query parameter.

app/views/components/gds/gds_main_template.scala.html
MaciejRewera: Good point. Didn't think about it this way.

app/models/SignOutReason.scala
SteveSugden: Yes, I know the benefits of enums.  I was just commenting that they are a pain to use in our services.  I'm fairly sure that using Java and SpringBoot you could just define the enum in a couple of lines and use it as a url param without any of this bind/unbind malarky.

app/views/components/gds/gds_main_template.scala.html
SteveSugden: I went down this path with timeout dialog.  Had a default param that all pages would have the dialog unless they said they didn't need it because they were used when the user was not logged in.

app/views/components/gds/siteHeader.scala.html
MaciejRewera: It seems so.

SteveSugden: That's good.  Do we have an opinion from Product Team about putting this feature in if it doesn't fully work?

app/assets/stylesheets/customsdecexfrontend-app.scss
PauloPortugal: That was my fault :sweat_smile:

app/views/components/gds/siteHeader.scala.html
```diff
@@ -14,17 +14,33 @@
...
+    language.Cy
+else
+    language.En
+}
+
+@signOutHref = @{
```
PauloPortugal, line 33: This is clever!

conf/messages
```diff
@@ -438,9 +439,12 @@ ileQueryResponse.timeout.message = Try again later.
...
-sessionTimout.back.button = Back to GOV.UK
 
 timeoutDialog.title = Your session will timeout soon
 timeoutDialog.message = For your security, we will sign you out in
 timeoutDialog.keepAlive.text= Stay signed in
-timeoutDialog.signout.text= Sign out 
\ No newline at end of file
+timeoutDialog.signout.text= Sign out
+
+userSignedOut.title = We have saved your answers and signed you out
```
PauloPortugal, line 448: This probably does not apply to Movements.

PauloPortugal: Ignore. This. Product team will have to look at this...

conf/messages
```diff
@@ -438,9 +439,12 @@ ileQueryResponse.timeout.message = Try again later.
...
+userSignedOut.information = To continue the declaration later, save this link:
+userSignedOut.startPageLink = https://www.tax.service.gov.uk/customs-movements/start
```
PauloPortugal, line 450: You need to remove the production domain @MaciejRewera

SteveSugden: Would it not be better to use the url defined in application.conf?
`urls.loginContinue`

PauloPortugal:
```
userSignedOut.startPageLink = {0}/customs-declare-exports/start
```
then pass the `@request.host` form the view.
This is a solution.

MaciejRewera: 👍

PR #277, opened 2020-06-18
CEDS-2481 Service Name link should point to FindConsignment page

PauloPortugal:


PauloPortugal:
> IMHO this should depend on feature switch for ILE Query. If it is disabled, we don't have this page at all.

Ahhhh!!!! Good point @MaciejRewera . completely forgot about the feature toggle.
So point to **Find Consignment** page is feature toggle enabled otherwise **Choice** page.

MaciejRewera:
I am not sure how this would work, because there is that `ReturnToStartException`. But I'd rather expect it to return 404 (Not Found) and default error page.

SteveSugden:
Yes,  It would have been nice if ILE Query endpoint re-directed to Choice if not enabled.  But I think someone wanted an "error" to obscure the end-point.  Can't remember.

PauloPortugal:
@MaciejRewera @SteveSugden Could you please review this again?

PR #282, opened 2020-07-09
CEDS-2609 - workaround for an issue with govukSummaryList 

SteveSugden:
The SummaryList component renders an empty <span> element when no "actions" (change links) are required _if_ there are any rows in the List that _do_ have an action.

So the workaround is to split the two rows into two SummaryLists and add a css class on the first to reduce the space between them.

Visually, the two single-row lists look the same as the original.

### DIFF
--- a/app/views/associateucr/associate_ucr_summary_no_change.scala.html
+++ b/app/views/associateucr/associate_ucr_summary_no_change.scala.html
@@ -22,12 +22,12 @@
 @import forms.ManageMucrChoice
 
 @this(
-    govukLayout: gds_main_template,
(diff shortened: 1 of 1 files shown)


app/views/associateucr/associate_ucr_summary_no_change.scala.html
```diff
@@ -64,7 +64,13 @@
                         content = Text(consignmentRef)
                     ),
                     actions = None
-                ),
+                )
+            ),
+            classes = "govuk-!-margin-bottom-1"
```
PauloPortugal, line 69: What is the issue?

PR #284, opened 2020-07-09
CEDS-2564 - remove un-used entries in messages resource file

SteveSugden:


SteveSugden:
Could you take a second look at this @PatrykRudnicki and @MaciejRewera ?   I removed UnitViewSpec as suggested.

test/unit/config/AppConfigSpec.scala
```diff
@@ -225,3 +226,7 @@ class AppConfigSpec extends WordSpec with MustMatchers with MockitoSugar {
   }
 
 }
+
+object AppConfigSpec extends Injector {
+  val realAppConfig = instanceOf[AppConfig]
+}
```
PatrykRudnicki, line 232: What is for? You're not importing this anywhere. Is it necessary?

PatrykRudnicki: Okay, I see the usage of this. I'm not sure if this should be in AppConfiSpec model.

My propose is to create a separate trait `WithAppConfig` or something like this and there prepare a correct app config. With this approach it will be just extends this trait and use injected AppConfig

SteveSugden: I just pulled that out of the deleted UnitViewSpec object.

PR #286, opened 2020-07-14
CEDS-2577 Introduce DUCR Part

MaciejRewera:
A few words of explanation for this PR:

- DUCR Part Details page is supposed to be visited instead of ILE Query pages.
- It is controlled by a feature switch.
- DUCR Part is treated the same way as DUCR, so all journeys required some
  changes to handle all 3 UCR types correctly.


app/forms/DisassociateUcr.scala
```diff
@@ -16,14 +16,17 @@
...
+import forms.DucrPartDetails.Separator
+import forms.UcrType.{Ducr, DucrPart, Mucr}
 import models.{ReturnToStartException, UcrBlock}
 import play.api.data.Forms._
 import play.api.data.{Form, Forms}
 import play.api.libs.json._
 import uk.gov.voa.play.form.ConditionalMappings._
 import utils.validators.forms.FieldValidator._
 
+import scala.util.Try
```
SteveSugden, line 28: Unused import?

hs_err_pid9425.log
```diff
@@ -0,0 +1,691 @@
+#
```
SteveSugden, line 1: This should be git-ignored.  And .DS_Store above

app/controllers/consolidations/ManageMucrController.scala
```diff
@@ -44,7 +44,7 @@ class ManageMucrController @Inject()(
 
   def displayPage(): Action[AnyContent] = (authenticate andThen ileQueryFeatureEnabled andThen getJourney(JourneyType.ASSOCIATE_UCR)) {
     implicit request =>
-      if (request.cache.queryUcr.map(_.is(UcrType.Ducr)).getOrElse(throw InvalidFeatureStateException)) {
+      if (request.cache.queryUcr.map(!_.is(UcrType.Mucr)).getOrElse(throw InvalidFeatureStateException)) {
```
PatrykRudnicki, line 47: Probably will be easier to read when is method could take more than one UcrType and we can invoke it `is(Ducr, DucrPart)`

hs_err_pid9425.log
MaciejRewera: 👍 Thanks! Got some issues with resolving conflicts so maybe this is where it came from.

app/utils/validators/forms/FieldValidator.scala
```diff
@@ -122,4 +113,16 @@ object FieldValidator {
   val areAllElementsUnique: Iterable[_] => Boolean = (input: Iterable[_]) => input.toSet.size == input.size
 
   val ofPattern: String => String => Boolean = (pattern: String) => (input: String) => input.matches(pattern)
+
+  val validDucr: String => Boolean = (input: String) => input.matches("""[0-9][A-Z][A-Z][0-9A-Z\(\)\-/]{6,32}""")
+
+  val validDucrIgnoreCase: String => Boolean = (input: String) => validDucr(input.toUpperCase)
```
PatrykRudnicki, line 119: Ah, I just realised that with previous ticket I should fix the DUCR validation in all services not only in one...

PatrykRudnicki: Hmm, can you use this validDucr method?

```
private val ducrPattern = Pattern.compile("[0-9]{1}[A-Za-z]{2}[0-9]{12}[-]{1}[-/()A-Za-z0-9]{1,19}")
val isValidDucr: String => Boolean = ducrPattern.matcher(_).matches()
```
This is from https://github.com/hmrc/customs-declare-exports-frontend/pull/1025/files#diff-787279139998872c746e28cb5b142d9bR169

app/controllers/consolidations/ManageMucrController.scala
MaciejRewera: Or maybe implement `isNot(_)` method instead?

PatrykRudnicki: Yeah, I thought about it, but is not is not saying what is allowed, not sure. `is(Ducr, DucrParts)` say more than `isNot(Mucr)`

In general this is just a suggestion, non blocking :)

MaciejRewera:
Sorry for the force push, but couldn't find other way to remove these unwanted files. I didn't see Patryk's comments, so I'll add 2 more commits.

app/utils/validators/forms/FieldValidator.scala
PatrykRudnicki: Or you can leave it, I can do this after you merge your PR, because this can break some tests, let's not combine thing together :)

PR #287, opened 2020-07-14
CEDS-2597 Fix DUCR validation

PatrykRudnicki:


test/unit/controllers/ileQuery/IleQueryControllerSpec.scala
```diff
@@ -140,7 +140,7 @@ class IleQueryControllerSpec extends ControllerLayerSpec with MockIleQueryCache
 
       "provided with correct DUCR" should {
 
-        val correctDucr = "9GB123456-QWERTY7890"
+        val correctDucr = "9GB123456789012-QWERTY7890"
```
MaciejRewera, line 143: Would it be worth updating `CommonTestData` ? There are couple DUCRs there.

PatrykRudnicki: Let me take a look, tests passed so I assume rest are correct

PatrykRudnicki: Yeah, CommonTestData has correct DUCR

PR #296, opened 2020-07-23
CEDS-2581 Fix for displaying DUCR parts information on submissions page

PatrykRudnicki:


conf/messages
```diff
@@ -274,8 +274,10 @@ submissions.shutmucr = Shut
 submissions.arrival = Arrive
 submissions.departure = Depart
 submissions.ducrassociation = Associate DUCR
+submissions.ducrpartassociation = Associate DUCR Part
```
SteveSugden, line 277: Are these messages not used anywhere?  Or are they created from the ucr type?

PatrykRudnicki: They're created from UCR type, that was missing or not added intentionally. Not sure about this, but right now the request type display that this is the Association for DUCR Part. It has been discussed with BAs about adding specific request type message for DUCR Type

PR #298, opened 2020-07-27
CEDS-2617 - Enter DUCR Part details when ILE Query is OFF

SteveSugden:


app/controllers/DucrPartChiefController.scala
```diff
@@ -0,0 +1,92 @@
...
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  def displayPage(): Action[AnyContent] =
+    (authenticate andThen isDucrPartsFeatureEnabled andThen ileQueryFeatureDisabled andThen getJourney(
+      JourneyType.ARRIVE,
+      JourneyType.DEPART,
+      JourneyType.ASSOCIATE_UCR,
+      JourneyType.DISSOCIATE_UCR
+    )) { implicit request =>
```
PatrykRudnicki, line 54: Just a suggestion. This can be extracted to the variable. You're using it in two places and it's quite big, few lines, but not sure about the name for it.

```
val requiredActions = (authenticate andThen isDucrPartsFeatureEnabled andThen ileQueryFeatureDisabled andThen getJourney(
      JourneyType.ARRIVE,
      JourneyType.DEPART,
      JourneyType.ASSOCIATE_UCR,
      JourneyType.DISSOCIATE_UCR
    ))
```

and you will be able to use it as `requiredActions { implicit request => .... }` or `requiredActions.async { implicit request => ... }`

app/views/components/config/ConsignmentReferenceConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+package views.components.config
+
+import config.DucrPartConfig
+import javax.inject.Inject
+import play.api.mvc.Call
+
+class ConsignmentReferenceConfig @Inject()(ducrPartConfig: DucrPartConfig) {
+
+  def backUrl: Call =
+    if (ducrPartConfig.isDucrPartsEnabled)
```
PatrykRudnicki, line 26: I'm not sure this if is enough.

query -> results -> choice -> consignment: back link to ChoiceController
DUCRPart from query -> choice -> Consignment: back link to ChoiceController
query disabled: choice -> DUCR Part -> consignment: back link DucrPart

So this if should looks like `if (ducrPartConfig.isDucrPartsEnabled && ileQueryDisabled)`. Don't know names of the config, but just use a pseudo code for it

app/views/components/config/DissociateUcrConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+package views.components.config
+
+import config.DucrPartConfig
+import javax.inject.Inject
+import play.api.mvc.Call
+
+class DissociateUcrConfig @Inject()(ducrPartConfig: DucrPartConfig) {
+
+  def backUrl: Call =
+    if (ducrPartConfig.isDucrPartsEnabled)
```
PatrykRudnicki, line 26: I'm not sure if this back link is correct, similar thoughts like above, that there is missing case, but not sure for what page this config is

app/views/components/config/ConsignmentReferenceConfig.scala
PatrykRudnicki: Ah, sorry, my fault, Consigment won't be presented when DUCR Part will be provided, so first two cases are wrong

app/views/components/config/DissociateUcrConfig.scala
PatrykRudnicki: So probably the same answer like above, it's okay :)

conf/app.routes
```diff
@@ -15,7 +15,11 @@ GET         /consignment-query/:ucr                controllers.ileQuery.IleQuery
 # DUCR Part details page
 GET         /ducr-part-details                     controllers.DucrPartDetailsController.displayPage()
 POST        /ducr-part-details                     controllers.DucrPartDetailsController.submitDucrPartDetails()
+POST        /ducr-part-details-journey             controllers.DucrPartDetailsController.submitDucrPartDetailsJourney()
 
+# DUCR Part CHIEF page
+GET         /ducr-part-created-chief               controllers.DucrPartChiefController.displayPage()
+POST        /ducr-part-created-chief               controllers.DucrPartChiefController.submit()
```
PatrykRudnicki, line 22: Are those urls from the ticket?

test/unit/views/disassociateucr/DisassociateUcrViewSpec.scala
```diff
@@ -16,26 +16,44 @@
...
 import play.twirl.api.Html
 import views.ViewSpec
 import views.html.disassociateucr.disassociate_ucr
 import views.tags.ViewTest
 
 @ViewTest
-class DisassociateUcrViewSpec extends ViewSpec with Injector {
+class DisassociateUcrViewSpec extends ViewSpec with MockitoSugar with BeforeAndAfterEach {
+
+  private val appConfig = mock[DucrPartConfig]
```
PatrykRudnicki, line 38: Probably better to use different name, because it's not AppConfig

test/unit/views/disassociateucr/DisassociateUcrViewSpec.scala
```diff
@@ -16,26 +16,44 @@
...
 
-  private val disassociatePage: disassociate_ucr = instanceOf[disassociate_ucr]
+  private val disassociatePage: disassociate_ucr = injector.instanceOf[disassociate_ucr]
 
   private def createView(form: Form[DisassociateUcr]): Html = disassociatePage(form)(request, messages)
 
+  override def beforeEach(): Unit = {
+    super.beforeEach()
+    when(appConfig.isDucrPartsEnabled).thenReturn(true)
+  }
```
PatrykRudnicki, line 50: You can remove `beforeEach` You're defining this behaviour in test cases.

In general it's not a good to define some behaviour two times, could cause some weird issues.

conf/app.routes
SteveSugden: Yes

app/controllers/DucrPartChiefController.scala
SteveSugden: Good idea

test/unit/views/disassociateucr/DisassociateUcrViewSpec.scala
```diff
@@ -64,7 +53,7 @@ class DisassociateUcrViewSpec extends ViewSpec with MockitoSugar with BeforeAndA
     }
 
     "display 'Back' button that links to Choice when ducrPart disabled" in {
-      when(appConfig.isDucrPartsEnabled).thenReturn(false)
+      when(config.isDucrPartsEnabled).thenReturn(false)
```
PatrykRudnicki, line 56: The `afterEach` part was necessary. Still you want to reset the mock after every test case

SteveSugden: Are you sure?

PatrykRudnicki: yeah, you should have defined the mock in every test case using method `beforeEach` if it's the same setting for everything or define it manually for every test case.

Every mock that is used should be reset in afterEach method, because reset should be always after every test case

SteveSugden: I know it's good practice and what we usually do.  But there is no real need to reset the mock every time.  Especially a simple mock of a boolean response.

So what would you like?   I originally was setting the mock behaviour in before.  Only 2 of the tests care what the mock returns - they set the mock manually.   The others work without setting the mock (as is currently the case).

PatrykRudnicki: Approved, I prefer in general the approach with afterEach etc methods, but it's not an issue in that simple mock, current approach works

PR #299, opened 2020-07-28
CEDS-2635 - Use Country dropdown in Movements Service

SteveSugden:


conf/messages
```diff
@@ -148,8 +148,8 @@ location.code.error = Goods location code is incorrect
 transport.title = Transport
 transport.heading = Depart consignment {0}
 transport.modeOfTransport.question = Which type of transport took the goods across the border?
-transport.modeOfTransport.empty = You need to choose mode of transport
-transport.modeOfTransport.error = Mode of transport is incorrect
+transport.modeOfTransport.empty = Select a mode of transport
+transport.modeOfTransport.error = Select a mode of transport
```
PatrykRudnicki, line 152: Are those messages intentionally the same? Should we keep one message?

The same question for transportId and nationality

SteveSugden: You can't really select an invalid value (it's a radio input).  But we do have lots of other examples of empty and wrong value having the same error message following content/error message review.    I tend to just leave them in-case someone decides to change the messages in future.

PatrykRudnicki: Yup, agree :)

PR #300, opened 2020-07-29
CEDS-2647 - Change time input to 12 hour time clock

SteveSugden:


app/forms/common/Time.scala
```diff
@@ -20,50 +20,60 @@ import java.text.DecimalFormat
...
+    def bind(hour: Try[Int], minutes: Try[Int], ampm: String): Time =
+      build(hour, minutes, ampm)
         .map(apply)
         .getOrElse(throw new IllegalArgumentException("Could not build time - missing one of parameters"))
 
-    def unbind(time: Time): (Try[Int], Try[Int]) =
-      (Try(time.time.getHour), Try(time.time.getMinute))
+    def unbind(time: Time): (Try[Int], Try[Int], String) = {
+      val timeString = time.time.format(time12HourFormatter)
+      (Try(timeString.split(":")(0).toInt), Try(time.time.getMinute), timeString.takeRight(2))
```
PatrykRudnicki, line 61: There is method `getHour`. Is it possible to use it on the first argument?

For second one you used `getMinute`.

Additionally I found method `get` in the `LocalTime` https://docs.oracle.com/javase/8/docs/api/java/time/LocalTime.html#get-java.time.temporal.TemporalField-
There you can specify what element from the time you want to retrieve. There is link to supported fields
https://docs.oracle.com/javase/8/docs/api/java/time/LocalTime.html#isSupported-java.time.temporal.TemporalField-
Some of the fields that could be useful here:
- HOUR_OF_AMPM
- CLOCK_HOUR_OF_AMPM (Not sure what's this, this will require some testing around)
- MINUTE_OF_HOUR
- AMPM_OF_DAY

SteveSugden: I was looking for something like that - getting the hour as a number from 1-12 and getting the AM/PM indictor.  In the end I just turned the time into a 12-hour format string, as you can see.   I'll take another look.

PR #307, opened 2020-08-08
CEDS-2089 - improved date/time error display

SteveSugden:
Includes
- highlighting of field level errors
- link to first field (day) in case of invalid date

### DIFF
--- a/app/views/arrival_details.scala.html
+++ b/app/views/arrival_details.scala.html
@@ -37,8 +37,10 @@
 
 @(form: Form[ArrivalDetails], consignmentReferenceValue: String)(implicit request: Request[_], messages: Messages)
 
-@removeDuplicateMessages(errors: Seq[FormError]) = @{
-    Map(errors map { a => a.message -> a }: _*).values.toSeq
+@processErrors(errors: Seq[FormError]) = @{
+  val removedDuplicates = Map(errors.reverse.map { a => a.message -> a }: _*).values.toSeq.reverse
+
+  removedDuplicates.map(err => err.copy(key = if(err.key == "dateOfArrival") "dateOfArrival.day" else err.key))
 }
 
 @govukLayout(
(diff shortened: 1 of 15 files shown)


app/views/arrival_details.scala.html
```diff
@@ -37,16 +37,24 @@
...
+@processErrors(errors: Seq[FormError]) = @{
+  val removeDuplicate = errors match {
+      case e1 :: e2 :: nil if e1.messages == e2.messages => Seq(e1)
+      case _ => errors
+  }
+  removeDuplicate.map(err => err.copy(key = err.key match {
+      case "dateOfArrival" =>  "dateOfArrival.day"
+      case "timeOfArrival" => "timeOfArrival.hour"
+      case _ => err.key
+  }))
```
MaciejRewera, line 49: How about this being a part of the form class? Or to build mapping in a way that no matter which field is incorrect, there is one and only one error message?

SteveSugden: We need 2 error message so that both the date and time fields are highlighted.  That was, I believe, the requirement for when the data&time are valid but either in the future or more than 60 days in the past.

What we don't want, however, is to display both error messages in the summary (because they are the same message - one with a date key and one with a time key)

test/unit/views/spec/ViewMatchers.scala
```diff
@@ -68,18 +68,14 @@ trait ViewMatchers { self: MustMatchers =>
 
   def haveChild(tag: String) = new ChildMatcherBuilder(tag)
 
-  def haveFieldError(fieldName: String, content: String): Matcher[Element] =
-    new ContainElementWithIDMatcher(s"error-message-$fieldName-input") and new ElementContainsFieldError(fieldName, content)
-
-  def haveFieldErrorLink(fieldName: String, link: String): Matcher[Element] = new ElementContainsFieldErrorLink(fieldName, link)
-
-  def haveGlobalErrorSummary: Matcher[Element] = new ContainElementWithIDMatcher("error-summary-heading")
-
```
MaciejRewera, line 77: 👍

app/views/arrival_details.scala.html
MaciejRewera: I see... I was only wondering if it would be possible to move this to the form class, so you don't have to write it every time you want to have the component on a page.

SteveSugden: Yes, I suppose I could extract that method do that the arrival and departure details forms can share it.

PR #308, opened 2020-08-12
CEDS-2669 - only process 'ACCEPTED' submissions

SteveSugden:


app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -47,7 +47,13 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
         case Success(response)  => logSuccessfulExchange("Submit Movement", response.body)
         case Failure(exception) => logFailedExchange("Submit Movement", exception)
       }
-      .map(_ => (): Unit)
+      .map(
+        response =>
+          response.status match {
+            case Status.ACCEPTED => (): Unit
+            case _               => throw new RuntimeException(s"Failed with response $response")
```
PatrykRudnicki, line 54: This is enough to handle it in the frontend probably, but I would use different type of Exception, because it's not a RuntimeException

SteveSugden: Yes.  RunTimeException was just my "see if it produces the right response on the UI" test.
I can maybe wrap that up in a function that can be applied to all the calls this connector makes.

PR #315, opened 2020-09-16
CEDS-2680 Update content for User Signed Out page

MaciejRewera:


app/views/user_signed_out.scala.html
```diff
@@ -32,15 +31,17 @@
     </p>
 }
 
+@paragraphBodyContent = {
```
PatrykRudnicki, line 34: Why you're not using `paragraphBody` component?

PatrykRudnicki (approved):

app/views/user_signed_out.scala.html
MaciejRewera: Because it only accepts `String` as input and I need to have link inside.

PR #319, opened 2020-09-29
CEDS-2689 Add hidden text with context for movements

janeturner:
- Create option for visually hidden text to link component
- Use link with hidden text on movements page
- Refactor files which use link component to use Html data type instead of String

### DIFF
--- a/app/views/movements.scala.html
+++ b/app/views/movements.scala.html
@@ -41,7 +41,7 @@
   } else {
     @link(
       message = components.elements_in_separate_lines(submission.ucrBlocks.map(_.fullUcr)),
-      textHidden = Some(messages("submissions.hidden.text", submission.ucrBlocks.map(_.fullUcr).mkString)),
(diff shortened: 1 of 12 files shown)


app/views/movements.scala.html
```diff
@@ -38,9 +39,11 @@
   @if(notifications.isEmpty){
     @components.elements_in_separate_lines(submission.ucrBlocks.map(_.fullUcr))
   } else {
-    <a href="@routes.NotificationsController.listOfNotifications(submission.conversationId)">
-      @components.elements_in_separate_lines(submission.ucrBlocks.map(_.fullUcr))
-    </a>
+    @link(
+      message = components.elements_in_separate_lines(submission.ucrBlocks.map(_.fullUcr)),
+      textHidden = Some(messages("submissions.hidden.text", submission.ucrBlocks.map(_.fullUcr).mkString)),
```
MaciejRewera, line 44: This is similar to the PR for Internal frontend. `mkString` without parameters will glue elements together.

MaciejRewera (approved):

PR #325, opened 2020-11-19
[CEDS-2745] Add feedback link to banner in all the pages

glidester:
As a part of Public Beta Readiness prep activities, every page
of the External Movements must have a feedback link at the
top left corner under black GOV.UK banner and next to BETA
blue box.

This sends the feedback to Splunk for later retrieval.

### DIFF
--- a/app/config/AppConfig.scala
+++ b/app/config/AppConfig.scala
@@ -41,7 +41,7 @@ class AppConfig @Inject()(
     .map(uri => MongockConfig(uri))
 
   private lazy val contactFrontendBaseUrl = servicesConfig.baseUrl("contact-frontend")
-  private lazy val contactFrontendServiceIdentifier = loadConfig("microservice.services.contact-frontend.serviceId")
+  lazy val contactFrontendServiceIdentifier = loadConfig("microservice.services.contact-frontend.serviceId")
 
   lazy val keyStoreSource: String = appName
   lazy val keyStoreUrl: String = servicesConfig.baseUrl("keystore")
--- a/app/config/AppConfig.scala
+++ b/app/config/AppConfig.scala
@@ -40,8 +40,8 @@ class AppConfig @Inject()(
     .getOptional[String]("mongodb.uri")
     .map(uri => MongockConfig(uri))
 
-  private lazy val contactBaseUrl = servicesConfig.baseUrl("contact-frontend")
-  private lazy val contactFrontendServiceIdentifier = loadConfig("microservice.services.contact-frontend.serviceIdentifier")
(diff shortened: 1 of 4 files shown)


conf/application.conf
```diff
@@ -103,6 +103,13 @@ microservice {
       ducrPart: enabled
       betaBanner: enabled
     }
+
+    contact-frontend {
+      protocol = http
+      host = localhost
+      port = 9250
+      serviceIdentifier = "customs-movements-frontend"
```
MaciejRewera, line 111: What do you think about naming this `serviceId`, as it is in contact-frontend? I am asking because I'd probably stay with the same name.

app/config/AppConfig.scala
```diff
@@ -40,6 +40,9 @@ class AppConfig @Inject()(
     .getOptional[String]("mongodb.uri")
     .map(uri => MongockConfig(uri))
 
+  private lazy val contactBaseUrl = servicesConfig.baseUrl("contact-frontend")
```
MaciejRewera, line 43: Could you rename it to `contactFrontendBaseUrl`? So it's in th esame convention as the field below.

test/unit/views/PhaseBannerSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+package views
+
+import base.Injector
+import play.api.test.FakeRequest
+import play.twirl.api.Html
+import views.html.components.gds.phaseBanner
+
+class PhaseBannerSpec extends ViewSpec with Injector {
+
+  private val page = instanceOf[phaseBanner]
```
MaciejRewera, line 26: Nit: This is not a page really ;-) I'd rather call it `banner` for e.g. But this is not a big deal.

MaciejRewera (approved):
LGTM
Just left a few comments with open questions. But nothing that would be blocking.

conf/application.conf
```diff
@@ -103,6 +103,13 @@ microservice {
       ducrPart: enabled
       betaBanner: enabled
     }
+
+    contact-frontend {
+      protocol = http
+      host = localhost
+      port = 9250
+      serviceId = "customs-movements-frontend"
```
MaciejRewera, line 111: It would be good to confirm before merging if this `serviceId` is fine.

MaciejRewera (approved):
LGTM

MaciejRewera (dismissed):

MaciejRewera (approved):
I may have missed tests for `AppConfig`. I know we spoke about it recently, but in my PR I realized it is good for documentation purposes.
But I am approving anyway and I'll leave it with you to decide.

PR #326, opened 2020-11-20
CEDS-2667 Add HMRCReportTechnicalssue component

PauloPortugal:
This component is a non-ajax implementation and is the recommended
approach by the HMRC Play UI team, since the previous "GetHelpLink" was
not accessible.

This component does not work locally and it should simply open a new
page. We have been told that this will work in all the other
environments.

### DIFF
--- a/app/views/components/gds/gds_main_template.scala.html
+++ b/app/views/components/gds/gds_main_template.scala.html
@@ -21,6 +21,9 @@
 @import components.gds.{phaseBanner, siteHeader}
 @import uk.gov.hmrc.govukfrontend.views.html.components._
 @import uk.gov.hmrc.govukfrontend.views.html.components.implicits._
+@import uk.gov.hmrc.hmrcfrontend.views.html.components.HmrcReportTechnicalIssue
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.reporttechnicalissue.ReportTechnicalIssue
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.language.{En, Cy}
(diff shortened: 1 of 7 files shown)


app/views/components/gds/gds_main_template.scala.html
```diff
@@ -63,6 +68,7 @@
 
   @if(useTimeoutDialog) {
       <script src='@routes.Assets.versioned("lib/hmrc-frontend/hmrc/all.js")'></script>
+      <script>window.HMRCFrontend.initAll();</script>
```
russellthorn, line 71: 👍

app/views/components/gds/gds_main_template.scala.html
```diff
@@ -89,6 +95,12 @@
         )))
     </div>
     @contentBlock
+    @hmrcReportTechnicalIssue(
+        ReportTechnicalIssue(
+            serviceCode = appConfig.keyStoreSource,
```
glidester, line 100: Nit: would be nicer if 'keyStoreSource' was renamed to 'appName' like in internal-frontend. Seems odd having a persistence var here (i know values are the same)

glidester (approved):

app/views/components/gds/gds_main_template.scala.html
PauloPortugal: @glidester your wish has come true.

PauloPortugal: Resolved

glidester (approved):

PR #329, opened 2020-12-10
CEDS-2771 Validate MUCR length is <36 characters long

PauloPortugal:
Recently we have encountered the issue with following MUCR:

     `GB/82F9-0N2F6500040010TO120P0A300689` (36 characters long)

It complies with our regex but it was over 35 characters long and ILE
rejected for being too long.
After speaking with the QAs we have determined that the regex requires
an overal length validation.

The approach taken was not to overload the existing regex, it is already
too complex to read and instead add a separate validation on it's
length.

### DIFF
--- a/test/unit/controllers/consolidations/MucrOptionsControllerSpec.scala
+++ b/test/unit/controllers/consolidations/MucrOptionsControllerSpec.scala
@@ -101,6 +101,15 @@ class MucrOptionsControllerSpec extends ControllerLayerSpec with MockCache with 
         status(result) mustBe BAD_REQUEST
         verify(page).apply(any(), any(), any())(any(), any())
       }
+
+      "a MUCR conforms with the regex but has been send but is over 35 characters long" in {
+        val incorrectForm = Json.toJson(MucrOptions("GB/82F9-0N2F6500040010TO120P0A300689", "GB/82F9-0N2F6500040010TO120P0A300689", Create))
+
+        val result = controller(AssociateUcrAnswers()).save()(postRequest(incorrectForm))
+
+        status(result) mustBe BAD_REQUEST
+        verify(page).apply(any(), any(), any())(any(), any())
+      }
     }
 
     "return 303 (SEE_OTHER)" when {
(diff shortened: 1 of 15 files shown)


MaciejRewera (commented):
LGTM, but what about `validMucrIgnoreCase` method? It is used in 5 other pages where user may want to enter MUCR.

The other question is whether `MucrOptions` could use `validMucrIgnoreCase` as well?

PauloPortugal:
@MaciejRewera please review again and thank you for spotting the missing code :)

PauloPortugal:
> The other thing is that MucrOptions forms are tested indirectly with MucrOptionsControllerSpec. It should not be a responsible of the controller tests to test the form class.

This is actually how the `MurcOptions` has been implemented. The `mapping` actually accepts anything passed through, even empty values, as the validation depends on the operation. So I am going to keep that as it is.

> It is a bit odd that `AssociateUcr` [...] has integration tests [...]
This has been addressed and the tests moved to the unit test. I never liked us moving these into `unit/it` folders. 
`AssociateUcrSpec` and `AssociateUcrItSpec` should have been more clearer.
Completely agree with you and I should have been more attentive ;)!

> It is a bit odd that [...] and `ShutMucr` have integration tests added but no unit tests
Completely agree with you and I should have been more attentive ;)
This has also been addressed.

test/unit/controllers/consolidations/MucrOptionsControllerSpec.scala
```diff
@@ -101,6 +101,15 @@ class MucrOptionsControllerSpec extends ControllerLayerSpec with MockCache with
         status(result) mustBe BAD_REQUEST
         verify(page).apply(any(), any(), any())(any(), any())
       }
+
+      "a MUCR conforms with the regex but has been send but is over 35 characters long" in {
```
MaciejRewera, line 105: Nit: This test name seems not being linguisticly correct. Although one should be able to tell what is it about.

MaciejRewera (dismissed):
LGTM
I sti ll don't like the approach for testing `MucrOptions`, but I understand it wasn't you starting it. It might be better to keep being concise before changing it totally.

test/unit/controllers/consolidations/MucrOptionsControllerSpec.scala
PauloPortugal: let me fix that after the current meeting :)

PauloPortugal: This has been addressed on e0ed4ff7f88c32b3b67621212cefce3796005359

MaciejRewera (approved):

MaciejRewera (approved):

PR #338, opened 2021-02-16
CEDS-2965 Set Root Url to redirect to home page

janeturner:
Create RootController to redirect to Choice page.
Remove references to 'Start' page and remove StartController.

### DIFF
--- a/app/controllers/RootController.scala
+++ b/app/controllers/RootController.scala
@@ -22,9 +22,9 @@ import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
 
 import javax.inject.Inject
 
-class StartController @Inject()(mcc: MessagesControllerComponents) extends FrontendController(mcc) with I18nSupport {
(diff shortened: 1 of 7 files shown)


app/controllers/RootController.scala
```diff
@@ -22,9 +22,9 @@ import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
 
 import javax.inject.Inject
 
-class StartController @Inject()(mcc: MessagesControllerComponents) extends FrontendController(mcc) with I18nSupport {
+class RootController @Inject()(mcc: MessagesControllerComponents) extends FrontendController(mcc) with I18nSupport {
 
-  def displayStartPage(): Action[AnyContent] = Action {
+  def displayPage(): Action[AnyContent] = Action {
```
glidester, line 27: NIT: doesn't actually display a page, as you're renaming it you could of picked something like 'redirectTo' :slightly_smiling_face:

glidester (approved):

PR #344, opened 2021-04-16
[CEDS-2947] Updating Validation Error Message - Yet more changes

glidester:


app/controllers/ConsignmentReferencesController.scala
```diff
@@ -44,12 +44,12 @@ class ConsignmentReferencesController @Inject()(
...
     implicit request =>
       val references = request.answersAs[MovementAnswers].consignmentReferences
-      Ok(consignmentReferencesPage(references.fold(form())(form().fill(_))))
+      Ok(consignmentReferencesPage(references.fold(form(request.answers.`type`))(form(request.answers.`type`).fill(_))))
   }
 
   def saveConsignmentReferences(): Action[AnyContent] =
     (authenticate andThen ileQueryFeatureDisabled andThen getJourney(JourneyType.ARRIVE, JourneyType.DEPART)).async { implicit request =>
-      form()
+      form(request.answers.`type`)
```
udinobi, line 52: Worth something like
```
val frm = form(request.answers.`type`)
```
as class field?

udinobi (approved):

app/forms/ConsignmentReferences.scala
```diff
@@ -57,29 +58,29 @@ object ConsignmentReferences {
...
-        "reference",
-        Mucr.codeValue,
-        text()
-          .verifying("consignmentReferences.reference.mucrValue.empty", nonEmpty)
-          .verifying("consignmentReferences.reference.mucrValue.error", isEmpty or validMucrIgnoreCase)
-      )
-    )(form2Model)(model2Form)
-
-  def form(): Form[ConsignmentReferences] = Form(mapping)
+  def mapping(goodsDirection: JourneyType) =
```
MaciejRewera, line 64: Nit: Is the name of this parameter correct? The type is completely different.

test/unit/forms/ConsignmentReferencesSpec.scala
```diff
@@ -35,62 +39,64 @@ class ConsignmentReferencesSpec extends BaseSpec {
   "Consignment References mapping" should {
     "return errors for empty fields" in {
       val inputData = ConsignmentReferences("", "")
-      val errors = ConsignmentReferences.form().fillAndValidate(inputData).errors
+      val errors = ConsignmentReferences.form(arrive).fillAndValidate(inputData).errors
```
MaciejRewera, line 42: Instead of using either `arrive` or `depart` you could put these values into a collection and then run tests for each element of this collection. Similar amount of code, but test coverage is much better.

test/unit/views/ConsignmentReferenceViewSpec.scala
```diff
@@ -37,6 +37,8 @@ class ConsignmentReferenceViewSpec extends ViewSpec with MockitoSugar with Befor
 
   private val page = injector.instanceOf[consignment_references]
 
+  private val goodsDirection = JourneyType.ARRIVE
```
MaciejRewera, line 40: Nit: Similar question about the name of this field. Is it correct?

app/controllers/ConsignmentReferencesController.scala
glidester: request.answers.`type` is a dynamic value that can change with each call to the saveConsignmentReferences method, so I'm not sure how you could make this an immutable class field val?

app/forms/ConsignmentReferences.scala
glidester: The two journey types relevant to the ConsignmentReferences screen are ARRIVE & DEPART so I though it was appropriate to say they are directions of travel for the goods. Happy to accept better suggestions though :)

test/unit/views/ConsignmentReferenceViewSpec.scala
glidester: The two journey types relevant to the ConsignmentReferences screen are ARRIVE & DEPART so I though it was appropriate to say they are directions of travel for the goods. Happy to accept better suggestions though :)

test/unit/forms/ConsignmentReferencesSpec.scala
glidester: I did consider this but as the only affect of using one or the other value is that a different I18n message key is selected (rather than an actual change in the code's behaviour) it did warrant the effort.

app/controllers/ConsignmentReferencesController.scala
udinobi: :+1:

app/forms/ConsignmentReferences.scala
MaciejRewera: I am very caucious with introducing new nomenclature. Especially, since we have similar terminology elsewere (`goodsLocation` for e.g.). I'd simply call it `journeyType` because this is the term used in other places in our services. It confused me already, so it would most certainly confuse someone in the future.

PR #348, opened 2021-05-17
[CEDS-3165] Goods location content change

glidester:
Adding in new expander with links to help users
fill in the form.

Also removing lazy keywords from variables in
AppConfig that are populated with the loadConfig
method as these do not fail fast atm if the key-
value is missing in the conf.


app/views/location.scala.html
```diff
@@ -55,5 +57,7 @@
         )
 
         @govukButton(Button(content = Text(messages("site.continue"))))
+
+        @goodsLocationExpander()
```
MaciejRewera, line 61: I've looked at the ticket and in the screenshots the expander is before 'Continue' button.

conf/messages
```diff
@@ -163,11 +163,37 @@ departure.details.error.overdue = Date and time of departure cannot be more than
 departure.details.error.future = Date and time of departure cannot be in the future
 
 location.title = Location
-location.question = What is the code for the location of the goods when they are declared?
-location.hint = This is usually the port they are leaving from and is officially known as the Goods Location Code. For example, if they were at Felixstowe, it would be GBAUFXTFXTFXT.
+location.question = What is the goods location code?
+location.hint = <p>This is a code for the location where goods are presented to customs. If necessary the goods location code must be updated as part of the arrival process, when the goods arrive at that location.</p><p>Most goods location codes contain just letters. For example, for goods leaving from Felixstowe, it would be GBAUFXTFXTFXT.</p>
```
MaciejRewera, line 167: I don't think it is a good idea to put html tags into messages. This should be done rather as partials or components. The visibility of styling/structure of content is worse and you'd need to look into multiple places to figure that out.

udinobi: Indeed. Yyou can easily have two `@paragraphBody`s inside braces and assigned to a wrapping var.

app/views/location.scala.html
glidester: Ah yes, good spot :+1:

conf/messages
glidester: Ok, I've genericides the exportsInputText component to allow `Content` to be passed as the 'hint' value rather than a messages key. It is a bit frustrating having to add a whole other message key-value pair just to put a space between two lines of content! :wink:

conf/messages
```diff
@@ -163,11 +163,38 @@ departure.details.error.overdue = Date and time of departure cannot be more than
...
+location.question = What is the goods location code?
+location.hint.paragraph1 = This is a code for the location where goods are presented to customs. If necessary the goods location code must be updated as part of the arrival process, when the goods arrive at that location.
+location.hint.paragraph2 = Most goods location codes contain just letters. For example, for goods leaving from Felixstowe, it would be GBAUFXTFXTFXT.
 location.code.empty = Enter a goods location code
 location.code.error = Enter a goods location code in the correct format
 
+location.expander.title=Find the goods location code (CDS data element 5/23)
+location.expander.intro=Select the location type, and then look up the code (links open in a new tab).
+location.expander.paragraph1.title=RoRo ports for road transport
+location.expander.paragraph1.text={0}<br/>Download the Tariff supplement and search for appendix 16R.
```
MaciejRewera, line 175: There are still a couple lines where you're using `<br/>` tag. Is it possible to get rid of that the same way as you did for others?

MaciejRewera (dismissed):

conf/messages
glidester: Ok, I've implemented the same sort of fix for the `<br>` tags but personally I'm not a fan. Not sure what the benefit of not having the simple tags like `<p>` and `<br>` in the messages file is, but IMO the resulting verbose code is much harder to reason about (as opposed to having these simple tags in the messages file).

MaciejRewera (approved):

PR #357, opened 2022-03-23
CEDS-3823 incorrect form validation on ducr part number

glidester:


app/utils/validators/forms/FieldValidator.scala
```diff
@@ -136,6 +135,6 @@ object FieldValidator {
 
   val validMucrIgnoreCase: String => Boolean = (input: String) => validMucr(input.toUpperCase) && noLongerThan(35)(input)
 
-  val isValidDucrPartId: String => Boolean = (input: String) => lengthInRange(1)(3)(input) && input.forall(_.isDigit)
-
+  def validRegex: String = "^[0-9]{0,3}[A-Z]?$"
+  val isValidDucrPartId: String => Boolean = (input: String) => lengthInRange(1)(4)(input) && input.matches(validRegex)
```
udinobi, line 139: Do you really need both expressions, `lengthInRange(1)(4)(input)` and `input.matches(validRegex)`, now that you have the regex? Does not the regex expression suffice?

glidester: No because without the lengthInRange(1)(4)(input) the regex allows empty string (you can have '111' or 'A' or `111A' but not '')

PR #366, opened 2022-05-05
CEDS-3579 Correct and simplify date validation

magicaljellybeans:
Add test case for all three fields being empty and add explanatory comments

### DIFF
--- a/app/forms/common/Date.scala
+++ b/app/forms/common/Date.scala
@@ -55,58 +55,50 @@ object Date {
   val twoDigitFormatter = new DecimalFormat("00")
   val fourDigitFormatter = new DecimalFormat("0000")
 
-  def validate(day: Try[Int], month: Try[Int], year: Try[Int]): Boolean = build(day, month, year).isSuccess
-  def validateNonEmptyFields(day: String, month: String, year: String): Boolean = Seq(day, month, year).forall(_.nonEmpty)
-  def isAnyFieldNotEmpty(fields: Seq[String]): Condition = mapping => fields.exists(field => mapping.getOrElse(field, "").nonEmpty)
+  def isValidDate(day: String, month: String, year: String): Boolean =
+    if (isAnyFieldEmpty(Seq(day, month, year))) true
+    else Try(LocalDate.of(year.toInt, month.toInt, day.toInt)).isSuccess
+  def isAnyFieldNotEmpty(fields: Seq[String]): Boolean = fields.exists(_.nonEmpty)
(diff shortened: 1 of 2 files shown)


app/forms/common/Date.scala
```diff
@@ -55,58 +55,50 @@ object Date {
   val twoDigitFormatter = new DecimalFormat("00")
   val fourDigitFormatter = new DecimalFormat("0000")
 
-  def validate(day: Try[Int], month: Try[Int], year: Try[Int]): Boolean = build(day, month, year).isSuccess
-  def validateNonEmptyFields(day: String, month: String, year: String): Boolean = Seq(day, month, year).forall(_.nonEmpty)
-  def isAnyFieldNotEmpty(fields: Seq[String]): Condition = mapping => fields.exists(field => mapping.getOrElse(field, "").nonEmpty)
+  def isValidDate(day: String, month: String, year: String): Boolean =
```
glidester, line 58: Maybe rename to `isValidDateOrEmpty

app/forms/common/Date.scala
```diff
@@ -55,58 +55,50 @@ object Date {
...
+    if (isAnyFieldEmpty(Seq(day, month, year))) true
+    else Try(LocalDate.of(year.toInt, month.toInt, day.toInt)).isSuccess
+  def isAnyFieldNotEmpty(fields: Seq[String]): Boolean = fields.exists(_.nonEmpty)
```
glidester, line 61: Maybe rename to `isAnyFieldPopulated` (save the sort of double negative use of 'not' & 'empty' together)?

app/forms/common/Date.scala
```diff
@@ -55,58 +55,50 @@ object Date {
...
+  def isAnyFieldEmpty(fields: Seq[String]): Boolean = fields.exists(_.isEmpty)
+  def isAnyFieldNotEmptyCondition(fields: Seq[String]): Condition = mapping => fields.exists(field => mapping.getOrElse(field, "").nonEmpty)
 
   def mapping(prefix: String): Mapping[Date] =
     Forms
       .tuple(dayKey -> dayMapping(prefix), monthKey -> monthMapping(prefix), yearKey -> yearMapping(prefix))
```
glidester, line 67: Maybe add a comment above this line stating "these checks only fire if at least one of these three fields contains some value, else these checks are skipped)"

magicaljellybeans: That could be a bit misleading though because the individual field mappings do do some checking, even if it's just to see if the field is empty.

I've tried to explain what you've mentioned on lines 78-81 - don't know if you've seen that or I need to try and word it better?

glidester (approved):

PR #372, opened 2022-06-06
CEDS-3879 Add dynamic section header to ducr details page

magicaljellybeans:


app/models/requests/RequestWithAnswers.scala
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package models.requests
+
+import models.cache.Answers
+import play.api.mvc.Request
+
+trait RequestWithAnswers[A] extends Request[A] {
+  def eori: String
```
udinobi, line 23: Minor thing... could not it be a `val`? Instead of a `def`, I mean.

udinobi (approved):
Just a minor comment

udinobi (approved):

PR #387, opened 2022-08-05
CEDS-4049 Correct back naviagtion from associate ucr summary page

magicaljellybeans:


app/views/associateucr/associate_ucr_summary.scala.html
```diff
@@ -27,11 +29,11 @@
         linkContent: linkContent,
         formHelper: uk.gov.hmrc.govukfrontend.views.html.components.FormWithCSRF)
 
-@(associateUcr: AssociateUcr, mucr: String)(implicit request: Request[_], messages: Messages)
+@(associateUcr: AssociateUcr, mucr: String)(implicit request: JourneyRequest[_], messages: Messages)
 
 @govukLayout(
     title = Title("associate.ucr.summary.title"),
-    backButton = Some(BackButton(messages("site.back"), consolidations.routes.AssociateUcrController.displayPage)),
+    backButton = Some(BackButton(messages("site.back"), AssociateUcrSummaryController.backlink)),
```
udinobi, line 36: Uhm :thinking: May I say that I do not like too much the view calling a controller's method? Why not placing the method directly in the view? If you need to also use elsewhere the method a views/helpers is a better place imo

magicaljellybeans: I did it this way because it's somewhat analogous to the way we have `Navigator` abstracted out in exports UI.
But at the same time I didn't want to create a whole new class for one backlink, so I opted for an object in the controller.

I generally think our views are not deciding on navigation?

magicaljellybeans: I mean technically it's not the controller class but the companion object.

udinobi: In same way they do that. If you think, once the form is submitted they redirect to the controller.

udinobi: I do not understand... why you would need to create a new class for the backlink!?

magicaljellybeans: As in I didn't want to go to the length of creating a `Navigator` for movements frontend.

udinobi (approved):

PR #390, opened 2022-09-26
Ceds 4096 save buttons

darrylmurphy:


app/controllers/consolidations/MucrOptionsController.scala
```diff
@@ -48,23 +49,25 @@ class MucrOptionsController @Inject() (
...
     form
       .bindFromRequest()
       .fold(
         formWithErrors => Future.successful(BadRequest(buildPage(formWithErrors))),
         validForm => {
           val updatedAnswers = request.answersAs[AssociateUcrAnswers].copy(mucrOptions = Some(validForm))
-          cacheRepository.upsert(request.cache.update(updatedAnswers)).map { _ =>
-            if (request.cache.queryUcr.isDefined)
+          if (request.cache.queryUcr.isDefined)
+            cacheRepository.upsert(request.cache.update(updatedAnswers.copy(readyToSubmit = Some(true)))).map { _ =>
```
magicaljellybeans, line 63: Instead of having this logic (flipping `readyToSubmit`) in each controller that precedes a summary page, would it be more convenient to just have it in the controller for the summary pages themselves? ie, a check to see if the flag has been flipped and if not, doing so then and there.

I'm thinking that might reduce number of places there needs to be changes if a page gets added. (granted there's not loads of places to begin with :laughing:)

darrylmurphy: I think the difference is negligible? The SummaryController.displayPage has no invocation of upsert, whereas its easy to set the flag with the answers on submission of last page in a journey.

PR #393, opened 2022-10-06
CEDS-4121 Add details component to DUCR part details page

magicaljellybeans:


glidester (approved):

app/views/ducr_part_details.scala.html
```diff
@@ -65,6 +69,13 @@
           inputClasses = "govuk-input--width-3"
         )
 
+        @govukDetails(Details(
+            summary = Text(messages("ducrPartDetails.details.title")),
+            content = HtmlContent(messages("ducrPartDetails.details.text", link(
+                message = Html(messages("ducrPartDetails.details.text.link")),
+                href = Call("GET", config.ucrInformationUrl))))
```
udinobi, line 76: If I'm not wrong, the link should open in a new tab. In that case, should not the `target` be specified?

PR #401, opened 2022-10-31
Ceds 4180 welsh translation

darrylmurphy:


glidester (approved):
With 1 suggestion :+1:

test/it/LanguageSwitchControllerSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+        redirectLocation(result).value mustEqual fakeUrl
+
+        cookies(result).find(_.name == "PLAY_LANG").get.value mustEqual "cy"
+
+        application.stop()
+      }
+    }
+
+  }
+}
```
glidester, line 76: Nit: Would be nice if we also tested the behaviour when an unsupported language is chosen (like Japanese) that it defaults to English

udinobi (dismissed):

glidester (approved):
:+1:

PR #413, opened 2023-02-07
CEDS 4352 change feedback link

EloraIP:


udinobi:
The PR is OK. Before the approval, would you mind to update in `plugins.sbt` the versions of the 2 plugins as reported by the platops' bot?

PR #468, opened 2024-09-05
CEDS-5968: remove ducr-part-created-chief and ducr-part-details pages

kenneth-rayner:


glidester (changes requested):
All changes look good, but could you also remove the DURC Part pages from the docs/Customs Declare Exports Movements AutoComplete.js script file (just to clean that up). Thanks

glidester (approved):

PR #472, opened 2024-10-30
CEDS-6139 Remove any CHIEF-related text

udinobi:


app/forms/DucrPartChiefChoice.scala
```diff
@@ -16,29 +16,15 @@
...
-import play.api.data.{Form, Forms, Mapping}
 import play.api.libs.json.{Json, OFormat}
-import utils.validators.forms.FieldValidator.isContainedIn
 
 case class DucrPartChiefChoice(choice: String) {
-  def isDucrPart = choice == DucrPartChiefChoice.IsDucrPart
+  def isDucrPart: Boolean = choice == DucrPartChiefChoice.IsDucrPart
 }
 
 object DucrPartChiefChoice {
```
magicaljellybeans, line 25: Is this class and object still needed?

magicaljellybeans (approved):

app/forms/DucrPartChiefChoice.scala
udinobi: Yes, it is still part of the `Cache` case class, but I do not know if we can remove from there.

PR #496, opened 2026-04-02
Dcwl 4237 update

jowensoc:


app/forms/ManageMucrChoice.scala
```diff
@@ -31,11 +31,18 @@ object ManageMucrChoice {
 
   val allChoices = Seq(AssociateThisMucr, AssociateAnotherMucr)
 
+//  val mapping: Mapping[ManageMucrChoice] =
```
sergi-hmrc: Do we need this?

jowensoc: Nope, I'll delete it

app/controllers/ChoiceController.scala
```diff
@@ -57,7 +56,7 @@ class ChoiceController @Inject() (authenticate: AuthAction, cacheRepository: Cac
     try
       nextPage(Choice(choice))
     catch {
-      case e: IllegalArgumentException => Future.successful(NotFound)
+      case _: IllegalArgumentException => Future.successful(NotFound)
```
sergi-hmrc, line 59: I know this isn't part of this PR but Choise.apply should return an Either/Try and we should map/match it:

```
object Choice {

  def apply(input: String): Either[IllegalArgumentException, Choice]
```

and then:
```
Choice(choice) match {
 case Right(c) => nextPage(c)
 case _ => Future.successful(NotFound)
}
```

app/repositories/RepositoryOps.scala
```diff
@@ -82,9 +83,9 @@ trait RepositoryOps[T] {
   def removeEvery[V](keyId1: String, keyValue1: V, keyId2: String, keyValue2: V): Future[Unit] =
     collection.deleteMany(and(equal(keyId1, keyValue1), equal(keyId2, keyValue2))).toFuture().map(_ => ())
 
-  def size: Future[Long] = collection.countDocuments().toFuture()
+  def size: Future[Seq[Long]] = collection.countDocuments().toFuture()
```
sergi-hmrc, line 86: I don't think this method is used anywhere but shouldn't it keep the type signature? 
```
def size: Future[Long] = collection.countDocuments().toFuture().map(_.sum)
```

app/repositories/RepositoryOps.scala
```diff
@@ -62,6 +62,7 @@ trait RepositoryOps[T] {
         options = FindOneAndReplaceOptions().upsert(true).returnDocument(ReturnDocument.AFTER)
       )
       .toFuture()
+      .map(t => t.head)
```
sergi-hmrc, line 65: IMO:
FindOne returns an Option:
`def findOne[V](keyId: String, keyValue: V): Future[Option[T]] `
findOneAndReplace should also return an Option:
`.map(_.headOption)`

test/base/MockAuthConnector.scala
```diff
@@ -49,8 +50,8 @@ trait MockAuthConnector extends Stubs with Injector {
     when(eoriAllowListMock.allows(any())).thenReturn(true)
   }
 
-  def userWithoutEori(user: SignedInUser = newUser("")): Unit = {
-    when(authConnectorMock.authorise(any(), ArgumentMatchers.eq(allEnrolments))(any(), any())).thenThrow(InsufficientEnrolments())
-    when(eoriAllowListMock.allows(any())).thenReturn(true)
-  }
+//  def userWithoutEori(user: SignedInUser = newUser("")): Unit = {
```
sergi-hmrc: remove method

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -28,8 +28,8 @@ import models.notifications.ResponseType.ControlResponse
 import models.notifications.queries.DucrInfo
 import models.notifications.queries.IleQueryResponseExchangeData.SuccessfulResponseExchangeData
 import models.submissions.Submission
-import org.mockito.BDDMockito.given
-import org.mockito.MockitoSugar.mock
+import org.mockito.BDDMockito.`given`
```
sergi-hmrc, line 31: we could also do: 
```
import org.mockito.BDDMockito.`given` as givenA
...
givenA(call).willReturn(result)
```

test/views/TransportViewSpec.scala
```diff
@@ -30,7 +30,7 @@ class TransportViewSpec extends ViewSpec with Injector {
 
   private implicit val request: JourneyRequest[AnyContentAsEmpty.type] = journeyRequest(ArrivalAnswers())
 
-  private def createPage(implicit request: JourneyRequest[_] = request): Html = page(Transport.form, "some-reference")
+  private def createPage(implicit request: JourneyRequest[_]): Html = page(Transport.form, "some-reference")
```
sergi-hmrc, line 33: Prefer the next scala 3 syntax here:

```
private given JourneyRequest[AnyContentAsEmpty.type] = journeyRequest(ArrivalAnswers())

private def createPage(using request: JourneyRequest[_]): Html = page(Transport.form, "some-reference")
```

Then all tests will look like: 
```
    "have the correct title" in {
      createPage.getTitle must containMessage("transport.title")
    }
```

And for the last example you will just need: 
```
checkAllSaveButtonsAreDisplayed(createPage(using journeyRequest(ArrivalAnswers(readyToSubmit = Some(true)))))
``` 
Notice the key `using` to inject the implicit parameter

build.sbt
```diff
@@ -3,7 +3,7 @@ import uk.gov.hmrc.versioning.SbtGitVersioning.autoImport.majorVersion
 val appName = "customs-movements-frontend"
 
 ThisBuild / majorVersion := 0
-ThisBuild / scalaVersion := "2.13.16"
+ThisBuild / scalaVersion := "3.3.7"
```
sergi-hmrc, line 6: Should we also address the following warning as part of this PR? 

```
It looks like you may be using LibSass to compile your Sass. LibSass is deprecated and will not be supported by the next major version of GOV.UK Frontend. See https://sass-lang.com/libsass/ for more information. To silence this warning, update $govuk-suppressed-warnings with key: "libsass"
```

test/controllers/ControllerLayerSpec.scala
```diff
@@ -44,15 +44,27 @@ abstract class ControllerLayerSpec extends UnitSpec with BeforeAndAfterEach with
...
+
+  private def jsonToFormFields(js: JsValue, prefix: String = ""): Seq[(String, String)] = js match {
+    case JsObject(fields) =>
+      fields.toSeq.flatMap { case (key, value) =>
+        jsonToFormFields(value, if (prefix.isEmpty) key else s"$prefix.$key")
+      }
+    case JsString(s)  => Seq(prefix -> s)
+    case JsNumber(n)  => Seq(prefix -> n.toString)
+    case JsBoolean(b) => Seq(prefix -> b.toString)
+    case JsNull       => Seq.empty
```
sergi-hmrc, line 65: nitpick: you don't need this one

app/utils/HashingUtils.scala
```diff
@@ -32,6 +32,6 @@ object HashingUtils {
     hmac.init(secretSpec)
 
     val sig = hmac.doFinal(value.getBytes("UTF-8"))
-    DatatypeConverter.printHexBinary(sig)
+    new String(Hex.encodeHex(sig, false))
```
sergi-hmrc, line 35: I would avoid creating new Strings:
```
java.util.HexFormat.of().withUpperCase().formatHex(sig)
```

app/repositories/RepositoryOps.scala
```diff
@@ -82,9 +83,9 @@ trait RepositoryOps[T] {
   def removeEvery[V](keyId1: String, keyValue1: V, keyId2: String, keyValue2: V): Future[Unit] =
     collection.deleteMany(and(equal(keyId1, keyValue1), equal(keyId2, keyValue2))).toFuture().map(_ => ())
 
-  def size: Future[Long] = collection.countDocuments().toFuture()
+  def size: Future[Seq[Long]] = collection.countDocuments().toFuture()
 }
 
-sealed abstract class WriteError(message: String)
+sealed abstract class WriteError(val message: String)
```
sergi-hmrc, line 89: I would remove this abstract class as only has 1 child

app/models/cache/JourneyType.scala
```diff
@@ -23,5 +23,5 @@ object JourneyType extends Enumeration {
 
   val ARRIVE, DEPART, ASSOCIATE_UCR, DISSOCIATE_UCR, SHUT_MUCR, JOURNEY_NOT_SELECTED = Value
 
-  implicit val format: Format[JourneyType] = Format(Reads.enumNameReads(JourneyType), Writes.enumNameWrites)
+  implicit val format: Format[JourneyType] = Format(Reads.enumNameReads(JourneyType), Writes.enumNameWrites[this.type])
```
sergi-hmrc, line 26:
```
given Format[JourneyType] = Format(Reads.enumNameReads(JourneyType), Writes.enumNameWrites[this.type])
```

test/views/LocationViewSpec.scala
```diff
@@ -32,7 +32,7 @@ class LocationViewSpec extends ViewSpec with Injector {
 
   private implicit val request: JourneyRequest[AnyContentAsEmpty.type] = journeyRequest(ArrivalAnswers())
 
-  private def createView(implicit request: JourneyRequest[_] = request): Html = page(form, "some-reference", None)
+  private def createView(implicit request: JourneyRequest[_]): Html = page(form, "some-reference", None)
```
sergi-hmrc, line 35: prefer `using` over `implicit`

sergi-hmrc (approved):

PR #502, opened 2026-07-10
DCWL-4809: DAC_Page_Structure_Usability_07 - Two headings at level 1 with identical text

DavouJ:


app/views/consolidations/mucr_options.scala.html
```diff
@@ -51,49 +51,45 @@
         @errorSummary(form.errors)
 
         @sectionHeader(messages("mucrOptions.heading", ucrBlock.map(_.ucr).getOrElse("")))
+        <fieldset class="govuk-fieldset" @if(form.hasErrors) { aria-describedby="declarantExporter-error" }>
```
sergi-hmrc, line 54: should we use `@govukFieldset` instead of the html version?

sergi-hmrc: Example of use:
```
@govukFieldset(Fieldset(
    legend = Some(Legend(
      content = Text("What is your address?"),
      classes = "govuk-fieldset__legend--l",
      isPageHeading = true
    )),
    html = html
  ))

@html = {
@govukInput(Input(
    id = "address-line-1",
    name = "addressLine1",
    label = Label(
      content = Text("Address line 1")
    ),
    autocomplete = Some("address-line1")
  ))
@govukInput(Input(
    id = "address-line-2",
    name = "addressLine2",
    label = Label(
      content = Text("Address line 2 (optional)")
    ),
    autocomplete = Some("address-line2")
  ))
@govukInput(Input(
    id = "address-town",
    name = "addressTown",
    label = Label(
      content = Text("Town or city")
    ),
    classes = "govuk-!-width-two-thirds",
    autocomplete = Some("address-level2")
  ))
@govukInput(Input(
    id = "address-postcode",
    name = "addressPostcode",
    label = Label(
      content = Text("Postcode")
    ),
    classes = "govuk-input--width-10",
    autocomplete = Some("postal-code")
  ))
}
```

In our case `@html` would just be the body with radios and legend the content of `@pageTitle`

sergi-hmrc (approved):

